PR #14, opened 2021-03-11
Jean cocogoat

sh560:
I did it anyways.
Still not sure how to account for c1 hold damage increase. I know there's a conditional thing but its only when holding. Should I just add another section for hold dmg?
Same for c4 anemo res decrease. Depends on enemy res and other sources....
Also noticed there was this bit `{DisplayPercent(50, s, getTalentStatKey("elemental", s))}` in one of Keqing's conts. Should I add something similar for Jean's passive3?


frzyc:
oh shoot. you should rebase this on master branch. I didnt realize closing the branch would close your PR

PR #15, opened 2021-03-11
Rebase jean

sh560:
It really didn't like one of the commits. Had to cherry pick my commits from branch after pulling the changes down.

Copying my previous comments here:

> I did it anyways.
> Still not sure how to account for c1 hold damage increase. I know there's a conditional thing but its only when holding. Should I just add another section for hold dmg?
> Same for c4 anemo res decrease. Depends on enemy res and other sources....
> Also noticed there was this bit {DisplayPercent(50, s, getTalentStatKey("elemental", s))} in one of Keqing's conts. Should I add something similar for Jean's passive3?


frzyc:
Went through and updated missing content.

PR #29, opened 2021-03-14
Jean

sh560:
Jean's on-hit healing scales with the damage she deals with auto attacks. Adjusted the passive to display accordingly. 

Tested numbers from calc to in game and they match. Used lvl80/80 Jean with tlvl:6 for all talents.


sh560:
Wait i'm dumb

PR #37, opened 2021-03-20
[Xinyan] Add Stats, Skill, Burst, Passives and Constelations

Drizmi:
Still needs work


Drizmi:
Constelations and passives need to be fully implemented, I only made the description for them

PR #40, opened 2021-03-22
Xinyan

Drizmi:
I hope it is what I want


Drizmi:
I am not sure how to implement phys dmg burst (basic formula uses element)

Drizmi:
C2 isn't done, future characters might deal physical dmg with their skills and burst and the current getStatKey does not support that

PR #44, opened 2021-03-26
Added new action on filtered artifacts: Lock or unlock with 1 click

mdnpascual:
Add a 1 button click to quickly toggle lock or unlock status of filtered artifacts.
Ignores the 50 artifact limit so it will change all artifact that matches the filter.

The flow I'm trying to simplify: https://discord.com/channels/785153694478893126/785176548428087306/824859650254962707

Starting state:
image

After clicking the button:
image


frzyc:
Seems like a niche use case, and adds a confusing button that doesn't have an obvious use. 
there are features planned that will cater to your use-case without adding confusion.
For now, just equip the artifacts you want to compare and compare via generated URL in two tabs
[https://trello.com/c/WsWnw531/103-lock-slots-against-build-generation](url)
[https://trello.com/c/AKcsXZTU/75-save-multiple-artifact-builds-to-character](url)

mdnpascual:
Looks like I'm running my own local build now. Quickly comparing an artifact(s) to the current equipped set on a character is the ONLY feature missing compared to the other calculator.

https://trello.com/c/WsWnw531/103-lock-slots-against-build-generation was exactly the way I'm doing it but since the number of artifacts in my inventory is blowing up, I wanted to quickly lock all of them and unlock the artifacts I want to compare for the build  generator.

PR #51, opened 2021-04-10
perf: optimize main stats & sub stats ocr reader

pedox:
Sometimes tesseractjs gives output like

```
+ ATK+4,.2%   // double period or comma
 « ATK+4,2%   // non Alphabet character
3,.358        // also double period or comma in main stat
```
because that some other sub stat cannot be detected.

here before fixed

Screenshot 2021-04-10 101255

and here after fixed

Screenshot 2021-04-10 101244


frzyc:
I see here that your decimal eliminators are commas. Which language do you play in? I did not realize that was an option.

pedox:
yeah javascript / typescript, sometimes it gives comma sometimes it gives period

pedox:
i use Indonesian, sometimes it gives comma sometimes it gives period

frzyc:
how does a artifact main stat of like HP 4780 look in your game? can you include a screenshot?
for example, in my game, it is in the form of `4,780` in my game
GenshinImpact_xYT3SYvRWl

pedox:
here it its

image

LrkDev:
It depends on language english (us, uk) use "." as decimal separator and "," as  thousands separator
Conversely Europe and indonesia use "," for decimal and "." for thousands
I think they all can use space for thousands separator

So seems like Indonesian translators kept english but didn't switch the logic

I'm french, adding to the dot and comma thing, the key are differents so need to switch language to take screenshots

src/Artifact/UploadDisplay.jsx
```diff
@@ -597,12 +598,12 @@ function parseMainStatvalue(recognition, defVal = { mainStatValue: NaN }) {
   let texts = recognition?.data?.lines?.map(line => line.text)
   if (!texts) return defVal
   for (const text of texts) {
-    let regex = /(\d+\.\d)%/
+    let regex = /(\d+[,|\\.]+\d)%/
     let match = regex.exec(text)
-    if (match) return { mainStatValue: parseFloat(match[1]), unit: "%" }
+    if (match) return { mainStatValue: parseFloat(match[1].replace(/,/g,".").replace(/\.{2,}/g, ".")), unit: "%" }
     regex = /(\d+,\d{3}|\d{2,3})/
```
frzyc, line 604: this only works with `X,XXX` and `XXX`, but would not work for `X.XXX`

pedox:
@LrkDev yeah i just realized to since im playing in indonesia from begining, not paying attention to the separator

src/Artifact/UploadDisplay.jsx
pedox: whoops missed, fixed on last commit

PR #56, opened 2021-05-20
Eula C4 update

Jilanow:
-C4 is now an additive multiplier 
-C4 can be toggled
-C6 text in Q only shows up when C6 is unlocked
-update icons


frzyc:
Thanks a lot for the PR. The features are now implemented on master.

PR #57, opened 2021-05-24
Update char_noelle_gen.json

ninjamask:
missing space


frzyc:
This file is generated from datamined files, probably not a good idea for it to be versioned, since it will be replaced every time it gets generated.

PR #59, opened 2021-06-05
Geo traveler

sh560:
ITS READY!


sh560:
Changes include adding CA2 for female traveler and tests

PR #60, opened 2021-06-26
Implement reactionChance to specify how likely an elemental reaction is to happen

Kouzukii:
Hey!

First of all, great work, I really like your optimizer.

But what's always been bugging me is the available choices for elemental reactions.
Either you have no reaction which makes EM useless, or you have a 100% chance of melt/vaporize which makes the optimizer choose EM over all other stats.

<https://genshinimpactcalculator.com/genshinCalc> goes around this problem by allowing you to specify how likely an elemental reaction is to happen.

This PR is just a rough idea of how to implement it and I'd like to hear your thoughts on how it could be improved.

image

image

It should default to 100% reaction chance, so unless the reaction chance is changed it should continue to function like before.

### DIFF
--- a/src/Build/BuildDisplay.tsx
+++ b/src/Build/BuildDisplay.tsx
@@ -151,7 +151,7 @@ export default function BuildDisplay({ location: { characterKey: propCharacterKe
     else isMounted.current = true
   }, [characterKey, maxBuildsToShow])
 
-  //validate optimizationTarget 
+  //validate optimizationTarget
   useEffect(() => {
     if (!statsDisplayKeys) return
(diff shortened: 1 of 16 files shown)


Kouzukii:
I noticed on the optimizer window it looks a bit off, I'll optimize that tomorrow.

image

Kouzukii:
> I noticed on the optimizer window it looks a bit off, I'll optimize that tomorrow.
> 
> image

Changing "No Reactions" to "No Reaction" seems to solve the issue for now

frzyc:
Since we are migrating to a new system, this feature will no longer be required. Thanks for your PR, sorry we couldn't see this to fruition.

PR #65, opened 2021-08-16
Refine 3 Broken pines using incorrect atk value

lemonlust:
Refinement 3 of Broken Pines shows an incorrect bonus attack stat, despite having the correct description


frzyc:
Already fixed with 5.19.3

PR #89, opened 2022-01-12
OCR Testing

Jugbot:
**This PR adds unit testing to OCR.** 
- Input accepts an Artifact -> creates fake screenshot -> runs OCR -> compares the two while also giving a link to the image that was parsed.
- While the default test always passes, many other generated artifacts will fail (consistently) for seemingly random reasons. Therefore this test will be primarily for debugging purposes
- The background of the image was photoshopped by me and the text position/size was hardcoded
- I didn't bother to add the proper artifact name or artifact image (though imo this would only potentially make tests fail more)


.gitignore
```diff
@@ -41,3 +41,5 @@ _TEMPLATE
 # template files
 TEMPLATE*
 *_ignore.png
+# tesseract language file
+*.traineddata
```
Jugbot, line 45: When running locally tesseract.js caches the traindata as a file

src/Artifact/ScanningUtil.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+
+  test('default', async () => {
+    await expect(defaultSubject).toPassOCR()
+  })
+
+  describe.skip.each(allSlotKeys)('%s', (slotKey) => {
+    describe.each(ArtifactSlotsData[slotKey].stats)('%s', (mainStatKey) => {
+      test('Artifact', async () => {
+        const subject: IArtifact = {
```
Jugbot, line 172: About have of these fail for random reasons, say level being 9 instead of 8 and stuff like that 🤷

src/Artifact/ScanningUtil.tsx
```diff
@@ -18,7 +18,7 @@ const schedulers = new BorrowManager(async (language): Promise<Scheduler> => {
   const promises = Array(workerCount).fill(0).map(async _ => {
     const worker = createWorker({
       errorHandler: console.error
-    })
+    })
```
Jugbot, line 21: Have you thought about using https://prettier.io/docs/en/install.html

src/Artifact/ScanningUtil.tsx
```diff
@@ -111,9 +112,7 @@ function imageDataToCanvas(imageData: ImageData) {
   return canvas // produces a PNG file
 }
 
-async function ocr(imageURL: string): Promise<{ artifactSetTexts: string[], substatTexts: string[], whiteTexts: string[], rarities: Set<Rarity> }> {
-  const imageData = await urlToImageData(imageURL)
```
Jugbot, line 115: This utility does not work on node due to certain libraries being undefined

frzyc:
So our focus for GO has not been on the OCR for a while. the code here are more ducttaped together... 
Looks like you've gotten started on https://trello.com/c/7vzb4dNC/122-artifact-scanner-automation-testing .

Now, a few things comes to mind. Right now, we have very rudimentary post processing for artifacts. so rudimentary, that each artifacts are scanned 3 times for each different region that's been processed differently. OCR speed can be improved a lot if we can isolate each line of the artifact for scanning. histogram analysis is probably sufficient in this case.

Secondly, the current OCR implementation has a lot of problem distinguishing 1s from the text. This is an sideeffect of using the default english model, instead of a model trained on genshin font. There are some trained models floating around that are trained on this font, but I never gotten tesseract to work with this model. Perhaps newer version of tesseract.js can offer better support for external trained models.

Lastly, if we are to provide our own trained models, we will have to consider what it means for deployment. Currently, the bundled website is around 3-4mb. a trained model file is in the range of 5-12mb. this would more than double our website deployment size. Not sure if there is a way to "lazy" load this file when we are using github page hosting....

Ultimately, I commend your efforts in attempting to have some sort of testing set up for our OCR system. I always believe that the OCR system needs a pretty heavy rewrite before unit testing it makes sense. You take that how you will. However, my support to your efforts will be somewhat unlimited, since we are focusing on a pretty big rewrite of the formula engine, and OCR is unpriorized for now.

PR #95, opened 2022-01-31
Fixing typos

chweesee:


frzyc:
Will be fixed when waverider is finished. thank you for the PR.

PR #108, opened 2022-02-18
Add artifacts and fix Yae incorect charged attack name

mineminemine:
- Thundersoother
- Traveling Doctor
- Wanderer's Troupe
- Fix for Pale Flame set


src/Data/Artifacts/PaleFlame/index.tsx
```diff
@@ -1,23 +1,34 @@
...
 
 const key: ArtifactSetKey = "PaleFlame"
 
-const [condStatePath, condState] = cond(key, "state")
+const [condStackPath, condStack] = cond(key, "stacks")
 
-const set2 = greaterEq(input.artSet.PaleFlame, 2, lookup(condState, { "twostacks": percent(0.5) }, percent(0.25)))
-const set4 = greaterEq(input.artSet.PaleFlame, 4, lookup(condState, { "onestack": percent(0.09), "twostacks": percent(0.18) }, naught))
+const physical_dmg_info: Info = { key: "physical_dmg_" }
+const set2 = greaterEq(input.artSet.PaleFlame, 2, percent(0.25), physical_dmg_info)
```
frzyc, line 16: This set2 effect was working fine before, why change it to be more complicated?

mineminemine: The previous one changes the 2 set effect to 50% if the condition is at two stacks. It might confuse users when just looking at the 2 set effect I guess? Refer to images here: https://imgur.com/a/3VsaNe9
But if you think its fine then I guess its fine.

frzyc: I think the existing implementation is fine. its rarely the case that only the 4set is displayed.

lantua: Don't you need to guard "twostack" behind C4? It could linger after PF4 are removed. The old code doesn't seem to do that

PR #110, opened 2022-02-19
Add more artifacts

mineminemine:
- Martial Artist
- Retracing Bolide
- Tiny Miracle
- Maiden Beloved (the sheet can probably be further improved)
- Scholar

Also rearranged the artifacts key in consts,ts


nguyentvan7:
Minor, but for Martial Artist conditional name, you can use `st("afterUse.skill")` using `st` from `SheetUtil`

frzyc:
Ahhh, if you move code in common files around you gonna cause conflicts with other ppl's code

PR #112, opened 2022-02-20
Add Bennett and minor fixes

nguyentvan7:
Add Bennett
Add infusion strings to translate sheet, changed Kazuha and Ayaka to use the new strings
Fixed Kazuha C1 + Skill CD showing strangely
Fixed Kazuha C6 showing when it shouldn't
Fixed Yae Miko C4 showing when it shouldn't
Fixed Yun Jin C1 + Skill CD showing strangely

2/20: Fix Yae Miko C4 icon being incorrect
2/20: Fix Yae Miko A4 calculating on `premod.eleMas` instead of `total.eleMas`


src/Data/Characters/YunJin/index.tsx
```diff
@@ -172,7 +172,10 @@ const sheet: ICharacterSheet = {
         node: infoMut(dmgFormulas.skill.shield, { key: `char_${key}_gen:skill.skillParams.3` }),
       }, {
         text: tr("skill.skillParams.4"),
-        value: data => data.get(input.constellation).value < 1 ? `${datamine.skill.cd}s` : `${datamine.skill.cd}s - 18%`,
+        value: data => data.get(input.constellation).value >= 1
```
frzyc, line 175: I wonder if its easier for the end user to read `9s - 18%` or `7.38s` without showing how that is calculated...

nguyentvan7:
```suggestion
        value: data => data.get(input.constellation).value >= 1
          ? `${datamine.skill.cd} - 18% = ${datamine.skill.cd*(1-0.18)}`
          : `${datamine.skill.cd}`,
```
In screenshot below, (1) is current implementation.
(2) in screenshot: We could use a WR node to represent the CD, in which case the '?' will appear with the formula. We would need a workaround to add the "s" unit for a WR node display though.
(3) in screenshot: We could show the equation as part of the text
image

Bennett's formula for A1 and A4-in-ult will be:
`cd*(1 + -1*0.8*0.5)`, or `cd - 20% - 50%`, which could also be represented as `cd - 60%` (I checked in game, the cooldown factors get multiplied, not added).

Not sure if those calculations will be confusing for end users since it doesn't logically add up how 20% and 50% CD reduction will create 60% unless we represent them as CD factors (aka `1-(80%*50%)`), but maybe that will also be confusing

frzyc: I'd say 3rd option for now. I dont really like the idea to boggling WR with these weird calculations that doesn't contribute much

nguyentvan7: Changed it to 3rd option for Bennett, Yunjin and Kazuha

nguyentvan7:
Bennett's C6 should automatically apply itself (if the weapon type matches, and user has selected as being within the area of the burst). It shows a bit weirdly for Catalyst characters right now, I think this is caused by the `target` bug with `canShow`, so hopefully will resolve itself once that is done.

PR #114, opened 2022-02-21
Add conditionalHeader, description to talentTemplate, standardize sheets and bug fixes

nguyentvan7:
Add `conditionalHeader` to `talentTemplate`, since it is needed for every conditional anyways.
Add `description` to `talentTemplate` for `conditional`.
Standardized sheets to use `talentTemplate` when possible.

Fixed bugs:
- Add `conditionalHeader` to all conditionals.
- Kazuha C2 not showing as team buff.
- Make Kazuha A4 description only appear once.
- Fix Raiden A4 using `total` ER.
    - This is considered a team-buff and should not be affected by dynamic-scaling buffs
    - There was a discussion about this on the waverider-bugs channel as well, there might need to be more fixes for other skills and whether they should scale off of `total` or `premod`.
- Fix some conditionals appearing when they shouldn't.
- Fix typos for Kazuha.
- Add display for Yunjin C2.

TODOs:
There is a known bug with `canShow` when using ~a node~ `UIData` that relies on `target`. Once that bug is fixed, there ~are 2 lines~ is 1 line to uncomment to hide conditionals more appropriately. ~One for Kazuha and~ one for Sucrose. Search for "uncomment after \`target\` bug is fixed"
Edit: Nevermind, Kazuha one was just from using the wrong node for `canShow`

### DIFF
--- a/src/Data/Characters/YunJin/index.tsx
+++ b/src/Data/Characters/YunJin/index.tsx
@@ -185,8 +185,6 @@ const sheet: ICharacterSheet = {
         value: datamine.burst.enerCost,
       }], {
         teamBuff: true,
-        header: conditionalHeader("burst", tr, burst),
-        description: tr("burst.description"),
         value: condBurst,
         path: condBurstPath,
         name: trm("burst"),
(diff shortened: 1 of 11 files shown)


src/Data/Characters/YunJin/index.tsx
```diff
@@ -185,8 +185,6 @@ const sheet: ICharacterSheet = {
         value: datamine.burst.enerCost,
       }], {
         teamBuff: true,
-        header: conditionalHeader("burst", tr, burst),
-        description: tr("burst.description"),
```
frzyc, line 189: Why remove the header/desc here? YunJin's burst is a teambuff, which needs these fields to display, right?

PR #116, opened 2022-02-22
Fix Shenhe quills applying damage increase to everything; other minor fixes

nguyentvan7:
Fix Shenhe quills applying damage to everything (it should only apply to Cryo damage).
In the process, I also futureproofed the system a lot for any damage increase with the Cartesian product of the `element`x`move` sets. Everything in the game right now can be covered with just the addition of the `element` and `move` sets, rather than the Cartesian product, but figured protecting for the future can't hurt.

If you think this is too extreme, I can reduce it to the addition of the two sets.

I also changed her bonus damage to show with the Cryo colored text (since it can only be applied with Cryo damage)

EDIT: I've reduced the number of keys to just the union/addition of the `element` and `move` sets. No need to clutter up the keys if nothing uses it (yet...)


frzyc:
You are right, Shenhe is implemented incorrectly currently in WR, and your implementation would remedy this problem. However, we are apprehensive about introducing all these keys. I think one of our issues with the old system was the amount of key bloat as we try to monkey patch more and more of mhy's implementation. the more keys we add, the more complex the "core" formula becomes to maintain.

I foresee us doing a "key sweep" in the near future, after we finish all the sheets, to reduce some of the existing key bloat.(A lot of them were introduced to hook up old UI requirements).

In the case of Shenhe, it would make sense to wrap up `nodeSkill` in `match(input.hit.ele, "cryo", ...)` to account for the cryo interaction.

src/Data/Characters/Shenhe/index.tsx
```diff
@@ -156,7 +156,7 @@ export const data = dataObjForCharacterSheet(key, elementKey, "liyue", data_gen,
   },
   teamBuff: {
     premod: {
-      all_dmgInc: nodeSkill,
+      cryo_elemental_dmgInc: nodeSkill,
```
nguyentvan7, line 159:
```suggestion
      all_dmgInc: equal(input.hit.ele, "cryo", nodeSkill),
```

Tried this out, but wasn't working for me

frzyc: Looks like there is a bug where hit specific entries like `ele/crit/base` cant be read behind `teambuff`. should be fixed soonTM

frzyc:
After discussion with @lantua:
Updating the `DataNode` to accommodate reading of these attributes behind `teambuff` space is more difficult than originally anticipated. Although its not pretty to add all these keys, it might be more prudent to add the keys as originally intended in this PR, to unblock more sheets development and patch wonky numbers. Once the modifications to WR is done, refactor we can then refactor these keys away. Will have a final look this weekend and probably merge in this PR as is.

src/Data/Weapons/Catalyst/EverlastingMoonglow/index.tsx
```diff
@@ -13,11 +13,11 @@ const key: WeaponKey = "EverlastingMoonglow"
 const data_gen = data_gen_json as WeaponData
 const hp_conv = [0.01, 0.015, 0.02, 0.025, 0.03]
 const [, trm] = trans("weapon", key)
-const normal_dmgInc = prod(subscript(input.weapon.refineIndex, hp_conv, { key: '_' }), input.premod.hp)
```
frzyc, line 16: There is no need for this change, right? The description here says "Normal Attack DMG is increased...", it shouldn't increase dmg for all elemental DMG.

nguyentvan7: It's written that way to increase normal attack damage for all elements. Since we are worried about key bloat, and nothing in the game is specific to an element + move type combination, I'm just going to reduce the keys down. Then this will just turn back to what it was before

nguyentvan7:
> After discussion with @lantua: Updating the `DataNode` to accommodate reading of these attributes behind `teambuff` space is more difficult than originally anticipated. Although its not pretty to add all these keys, it might be more prudent to add the keys as originally intended in this PR, to unblock more sheets development and patch wonky numbers. Once the modifications to WR is done, refactor we can then refactor these keys away. Will have a final look this weekend and probably merge in this PR as is.

I also was coming to this same conclusion, but wasn't 100% confident in my answer since I am not experienced in the WR system (other than reading it for a week lol). Couldn't really think of a way to do this without causing infinite loops in the teamBuff calculation

nguyentvan7:
Tidied up some stuff, like Mistsplitter sheet, small language variations, coloring of elemental values, etc.

frzyc:
oh dear, conflicts over eula changes... you might want to separate out your PRs, instead of stuffing it all into this singular PR, otherwise one issue would block the whole PR.

nguyentvan7:
Noted, and resolved

public/locales/en/sheet.json
```diff
@@ -37,6 +37,8 @@
     "atk": "ATK Increase"
   },
   "skillCDRed": "Ele. Skill CD Reduction",
+  "pressCD": "Press CD",
+  "holdCD": "Hold CD",
```
frzyc, line 41: So these values are already piped in from pipeline using genshin's translation.
image

The values in `sheet.json` needs to be manually translated for every language. The `sheet_gen.json` are values we extract from the genshin translations.

Perhaps you are opting to go this way because in Genshin `Hold CD` is in the format of `CD (hold)`?

nguyentvan7: Yea I'm a bit anal when it comes to strings. Can revert this if you think it's unnecessary translation work. I also thought about just modifying the English translation as it is generated but didn't really explore it

PR #117, opened 2022-02-22
Add Kujou Sara

mineminemine:
With some comments in the sheet
Edit: forgot to mention here that electro_critDMG_ does not work as of now.


nguyentvan7:
`src/Formula/index.ts`
Changing `common.hit.dmg.critHit` and `common.hit.dmg.avgHit` to below fix looks to fix the `electro_critDMG_` issue (along with the other elements)

```
        critHit: sum(unit, total.critDMG_,
          lookup(hit.ele, objectKeyMap(allElements, element => total[`${element}_critDMG_`]), 0)),
        avgHit: sum(unit, prod(total.cappedCritRate, sum(total.critDMG_,
          lookup(hit.ele, objectKeyMap(allElements, element => total[`${element}_critDMG_`]), 0)))),
```

PR #125, opened 2022-03-04
Possible fix to critDmg_ and critRate_ bugs

nguyentvan7:
From my small scale testing, this fixes the issues with `{ele}_critDmg_` and `{move}_critRate_` nodes not working properly. It didn't break custom entered `critRate_` or `critDmg_` either. This would also fix one of the issues in #117 (Sara C6)

### DIFF
--- a/src/Formula/index.ts
+++ b/src/Formula/index.ts
@@ -150,11 +150,11 @@ const common: Data = {
           break
         case "critRate_":
           operands.push(percent(0.05, { key, prefix: "default" }),
-            lookup(hit.move, objectKeyMap(allMoves, move => customBonus[`${move}_critRate_`]), 0))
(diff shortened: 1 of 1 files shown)


src/Formula/index.ts
```diff
@@ -150,11 +150,11 @@ const common: Data = {
           break
         case "critRate_":
           operands.push(percent(0.05, { key, prefix: "default" }),
-            lookup(hit.move, objectKeyMap(allMoves, move => customBonus[`${move}_critRate_`]), 0))
+            lookup(hit.move, objectKeyMap(allMoves, move => premod[`${move}_critRate_`]), 0))
```
frzyc, line 153: this is right before the actual accumulation of stats right before dmg calc, so instead of `premod`, this should be `total`

PR #127, opened 2022-03-05
Add optional header for sections, add header to weapon passives

nguyentvan7:


nguyentvan7:
I forgot I created a draft PR for this already, and did a rebase earlier, which is why this commit history looks kinda wonky.

PR #133, opened 2022-03-11
Fix Noblesse Oblige teambuff not showing

mineminemine:


nguyentvan7:
Not able to check at the moment, but I think I tried this fix on my local branch as well. This would allow the user to stack NO4 team buff, which they shouldn't be allowed to do

mineminemine:
Oh right, I totally forgot about that

PR #142, opened 2022-03-14
Fix "0" nodes appearing in TeamBuffCard

nguyentvan7:
Same fix as #135, just forgot to apply it to the TeamBuffCard as well.

Maybe `TeamBuffCard()` and `TeamBuffDisplay()` should be merged at some point? They are almost identical.


frzyc:
those two UI should be merged, or refactored, was just waiting for resonance to be put in first.

PR #145, opened 2022-03-16
Beidou

Jomata:
image

image

image


src/Data/Characters/Beidou/index.tsx
```diff
@@ -1,185 +1,271 @@
...
+const [condC6Path, condC6] = cond(key, "Constellation6")
+const [condA4Path, condA4] = cond(key, "Ascension4")
+
+const nodeC3 = greaterEq(input.constellation, 3, 3)
+const nodeC5 = greaterEq(input.constellation, 5, 3)
+
+const skillDmgOneHit = datamine.skill.dmgBase.map((dmg,i) => dmg + datamine.skill.onHitDmgBonus[i])
+const skillDmgTwoHits = datamine.skill.dmgBase.map((dmg,i) => dmg + 2 * datamine.skill.onHitDmgBonus[i])
+
+const electroShieldStr_ = { "customBonus": { "shield_": percent(1.5) } }
```
frzyc, line 85: this would make it additive instead of multiplicative?

Jomata: It's adding the multipliers that then get multiplied in the damage formula

`skillDmgOneHit ` ends up being `[121.6 + 160 , 130.72 + 172 , 139.84 + 184]` etc
`skillDmgOneHit ` ends up being `[[121.6 + 2*160 , 130.72 + 2*172 , 139.84 + 2*184]` etc

The numbers also match current GO

Current:
image

Waverider:
image

750.7 == 206+2*272

Also just realized that Waverider is rounding the cooldown from 7.5 to 8

frzyc: My comment is towards the `electroShieldStr_ `, looks like someone already added elemental shields, so we can fix it in post.

there is a `fixed` field in the `field` that can fix rounding

PR #163, opened 2022-03-21
Minor weapon changes

mineminemine:
Minor changes on the conditional headers

### DIFF
--- a/src/Data/Weapons/Sword/SkywardBlade/index.tsx
+++ b/src/Data/Weapons/Sword/SkywardBlade/index.tsx
@@ -12,7 +12,7 @@ import icon from './Icon.png'
 
 const key: WeaponKey = "SkywardBlade"
 const data_gen = data_gen_json as WeaponData
-const [tr, trm] = trans("weapon", key)
+const [tr] = trans("weapon", key)
 
 const [condPassivePath, condPassive] = cond(key, "SkyPiercingMight")
 const atkSrc_ = [0.2, 0.25, 0.3, 0.35, 0.40]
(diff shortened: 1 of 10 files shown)


src/Data/Weapons/Sword/SkywardBlade/index.tsx
```diff
@@ -39,7 +39,7 @@ const sheet: IWeaponSheet = {
     conditional: {
       value: condPassive,
       path: condPassivePath,
-      header: conditionalHeader(tr, icon, iconAwaken, trm("might")),
```
frzyc, line 42: why are we changing the conditional name to just "Conditional"?

mineminemine: All the weapons have it as "Conditional" at the moment since I wasn't too sure on what we're going to name them so I just named it "Conditional" and conform all the weapons to it. Kinda felt like it looked more explanatory when naming them as "Conditional" rather than naming them under their ability/passive name. But I guess it's entirely up to you on how we're moving forward with this.
image
image

frzyc: I rather not use the word "Conditional" since its a technical word used in the system. I think its fine to use the passive name. I don't see this change as an improvement.

mineminemine: How about weapons that don't particularly have an ability name ie. Engulfing Lightning?
image

Should the conditional name just be left empty?

frzyc: That could just be "Timeless Dream: Eternal Stove" or empty

mineminemine: Okay I'll make changes on all the weapon and make a new PR

PR #168, opened 2022-03-22
Refactor Shenhe. Minor sheet tweaks

nguyentvan7:
Refactor Shenhe to `talentTemplate` and `sheetTemplate`
Update her "active character in field" buff to properly show in the team buff panel
Add extra fields to display for her constellations

For the extra fields, I don't know if this makes her sheet too busy. If you think it is a bit excessive, I can remove some of the fields
Talent card for skill gets a bit long:
image

Same with her card on the team buff page. Had to zoom out more than usual to capture it all, you can see the length compared to Bennett and Ganyu, a lot of it is from her skill description (and most of that isn't relevant to the team buff section)
image

### DIFF
--- a/public/locales/en/char_Shenhe.json
+++ b/public/locales/en/char_Shenhe.json
@@ -1,6 +1,6 @@
 {
   "quill": "Icy Quill",
-  "asc4Cond":"After Shenhe uses <strong>Spring Spirit Summoning</strong>",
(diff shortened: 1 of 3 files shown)


frzyc:
Honestly, teambuff UI probably needs to be redesigned. I'd probably prefer a horizontal layout for each teammate, instead of vertical. a lot of the time its becoming cumbersome to scroll down to see the full character, which isnt ideal.

PR #206, opened 2022-04-01
Team Infusion

frzyc:


src/Data/Characters/dataUtil.tsx
```diff
@@ -19,15 +19,15 @@ const infusion = stringPrio(
...
-  // Inferred Element
-  equalStr(input.weaponType, "catalyst", input.charEle),
-  lookup(input.hit.move, {
-    "normal": infusion, "charged": infusion, "plunging": infusion,
-    "skill": input.charEle, "burst": input.charEle,
-  }, undefined),
-  "physical",
-)
+const meleeHitEle = lookup(input.hit.move, {
+  "normal": infusion, "charged": infusion, "plunging": infusion,
```
nguyentvan7, line 23: I think this breaks auto attack physical inference when non infused

PR #219, opened 2022-04-06
Update ArtifactDisplay

nguyentvan7:
Change `ArtifactDisplay` in `CharacterCard` to display artifacts in one row.
Also shows rarity, level and main stat value. Can also hover the main stat value to get the main stat name. Placeholder icons are used when an artifact slot is missing

Venti is triple EM, which used to take 3 rows. Border indicates rarity. Placeholder icons are not properly centered because the SVGs are not aligned 
image

Also added a couple tooltips. One for the inventory icon in `ArtifactCardNano` and one for the "# of threads" dropdown to explain what the option does.
image
image

Also added an "Exclude Artifacts" button in the Character > Artifacts page, and in `ArtifactBuildDisplayItem`. If you don't agree with this button, I can remove it, but it is pretty handy.
image
image

### DIFF
--- a/src/Components/Artifact/ArtifactCardNano.tsx
+++ b/src/Components/Artifact/ArtifactCardNano.tsx
@@ -85,5 +85,5 @@ function SubstatDisplay({ stat }: { stat: ICachedSubstat }) {
 }
 function LocationIcon({ location }) {
   const characterSheet = usePromise(CharacterSheet.get(location ?? ""), [location])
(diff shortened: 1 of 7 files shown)


nguyentvan7:
image
Hm, I tightened the spacing as much as I feel comfortable with, and it still will break in 3 column layout with 3 % artifacts equipped

nguyentvan7:
I made a change to make it a 3x2 grid and also show the set effects. I don't know how I feel about it. It solves the scaling issues, but it feels pretty clunky
image

nguyentvan7:
In case you didn't have enough choices, how about another one? Compact 5x1 grid with no scaling issues. Remove the main stat value, and put it in the tooltip
image

PR #222, opened 2022-04-06
Fix Build Details button being disabled on equipped builds

AleXu224:


frzyc:
This is intended. Showing details for an equipped build is disabled to facilitate future UI changes to character UI.

PR #223, opened 2022-04-07
Add tooltips

nguyentvan7:
Added a couple tooltips. One for the inventory icon in `ArtifactCardNano` and one for the "# of threads" dropdown to explain what the option does.
image
image

Below is now removed, keeping text for history-keeping
~~Also added an "Exclude Artifacts" button in the Character > Artifacts page, and in `ArtifactBuildDisplayItem`. If you don't agree with this button, I can remove it, but it is pretty handy.
image
image~~

### DIFF
--- a/src/Components/Artifact/ArtifactCardNano.tsx
+++ b/src/Components/Artifact/ArtifactCardNano.tsx
@@ -85,5 +85,5 @@ function SubstatDisplay({ stat }: { stat: ICachedSubstat }) {
 }
 function LocationIcon({ location }) {
   const characterSheet = usePromise(CharacterSheet.get(location ?? ""), [location])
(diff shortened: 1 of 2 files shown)


src/PageCharacter/CharacterDisplay/CharacterArtifactPane.tsx
```diff
@@ -57,6 +65,7 @@ function CharacterArtifactPane({ newBuild = false }: { newBuild?: boolean }) {
         <Grid container spacing={1}>
           <Grid item>
             {newBuild ? <Button onClick={equipArts} className="mr-2">Equip artifacts</Button> : <Button color="error" onClick={unequipArts}>Unequip all artifacts</Button>}
+            <Button onClick={excludeArts} sx={{ ml: 1 }}>Exclude all artifacts</Button>
```
frzyc, line 68: button should be disabled if all artifacts are all excluded. use `variant: info`

frzyc: actually, lets not do this exclusion button at all. I think this exclusion thing is something people are not using correctly. 
Once https://trello.com/c/X9og8lzc/180-priority-queue-for-use-equipped-build-option is added, this feature will be confusing to have.

nguyentvan7:
Ready for review again

PR #225, opened 2022-04-09
Fix Noelle/Yunjin shield, fix default artifact conditional counter

nguyentvan7:
Conditional selector will now delete the path if it is being set to `undefined`, so we can tell which artifact conditionals are actually set
Make Yunjin and Noelle use geo shields with 1.5x absorption
Fix constellation level text


src/PageBuild/Components/ArtifactConditionalCard.tsx
```diff
@@ -18,8 +18,10 @@ export default function ArtifactConditionalCard({ disabled }: { disabled?: boole
   const [open, setOpen] = useState(false)
   const onOpen = useCallback(() => setOpen(true), [setOpen])
   const onClose = useCallback(() => setOpen(false), [setOpen])
+  console.log(character.conditional)
```
frzyc, line 21: log

PR #228, opened 2022-04-14
Update artifact display, add teammate display

nguyentvan7:
image
image

New scaling demo: https://youtu.be/ZjMqw_P0tFI

One thing I forgot to show in the video is hovering the main stat icon for the value
chrome_nyQzUWgZ5Q

1 small issue
The side profiles are a bit low resolution when scaled up, since they are 128x128 (vs the 256x256 of the full portrait)
image

In team buffs, I've opted to keep the original weapon card to make better use of the space
image

Before/after comparison GIFs of the height of the card: https://discord.com/channels/785153694478893126/961187969736265778/964342545398267915


nguyentvan7:
Made the equipment/team stuff take 2 rows instead of 3. Also updated the scaling demo video and the screenshots with the changes

PR #239, opened 2022-04-16
Add ability to "favorite" characters

nguyentvan7:
Added new GO-only `boolean` property to `ICharacter`: `favorite`
Any character that has `favorite === true` will be displayed first in a list of characters, while also respecting the sort the user has chosen.
Most of this is done by filtering the list of characters twice, once with favorites, and once without. Then sorting those as needed, then combining.

`CharacterInventory` - The heart is clickable, and will trigger a resort of the inventory
image

`CharacterOverviewPane` - The heart is clickable
image

`CharacterSelectionModal` - The heart is clickable, and will trigger a resort of the modal
image

`CharacterDropdownButton` - The heart is not clickable
image


frzyc:
Looks like the button is very finnicky to click. I understand you did some hacks to make the favorite button clickable over the character card's larger clickable area. I think the interactable of this UI element should not overlap. The solution might be just make the whole card not clickable, and have a "details" button or something to go to the character details.

nguyentvan7:
> Looks like the button is very finnicky to click. I understand you did some hacks to make the favorite button clickable over the character card's larger clickable area. I think the interactable of this UI element should not overlap. The solution might be just make the whole card not clickable, and have a "details" button or something to go to the character details.
> 
> The "delete" button can possible be leveraged into this redesign.

The heart button being finicky was just a zIndex issue. If you think the click targets of the card still need to be reworked, I can work on that. It was one of the things I planned to do, but was going to wait until after 8.2.0 for it
I also wouldn't say the things I did are "hacks", it's just how it has to work when there are nested onClicks

PR #241, opened 2022-04-17
Fix some bugs, small refactor for reaction display

### LINKED ISSUE #230: Kokomi Burst's initial Damage scales with ATK instead Max HP
image

### LINKED ISSUE #238: [Bug] Wrong formulas RES for Max Sea-Dyed Foam DMG
The formulas for RES here should be **1-(RES/2)** with RES at -20%, missing the parenthesis.

And also OHC is independent of enemy def. (from keqing main https://library.keqingmains.com/equipment/artifacts#ocean-hued-clam)
So there is no DEF multiplier in the DMG formulas.

res

nguyentvan7:
Fix Kokomi burst scaling - Fixes #230
Fix Kokomi charged attack text - Fixes #230
Fix OHC damage calculation - Fixes #238
Fix blank formula cards showing in formula dropdown
Fix Thundering Fury 4p not showing all fields, and showing incorrectly in formulas
Refactor reaction fields to use NodeFieldDisplay
- With this change, we lose the dropdown for Crystallize+element and Swirl+element. And we also lose the elemental icons, but I feel like those weren't necessary
image
It gets more bulky in worst-case scenario, if that is a dealbreaker
image

Fix Chongyun A1 always applying,
Fix some Chongyun strings

### DIFF
--- a/public/locales/en/char_Chongyun.json
+++ b/public/locales/en/char_Chongyun.json
@@ -1,8 +1,8 @@
 {
-  "activeCharField": "Sword, Claymore, and Polearm wielding characters within the Frost Field",
   "infusion": "Sword, Claymore and Polearm Cryo Infusion",
   "passive2": "Summoned Sword DMG",
-  "asc4Cond": "Opponents hit by Rimechase Blade",
-  "constellation1":"Ice Blade DMG",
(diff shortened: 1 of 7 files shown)


nguyentvan7:
Fixes https://github.com/frzyc/genshin-optimizer/issues/238

PR #243, opened 2022-04-17
Fix compile warnings

nguyentvan7:
These are from my PR, my b


frzyc:
I've already fixed them.

PR #294, opened 2022-04-22
Fix resonance tooltip, adjust conditional tooltip

### LINKED ISSUE #233: Geo Infusion has empty popup
image

nguyentvan7:
Didn't mean to publish this branch on the main repo, woops
Fix #233 - Move tooltip to the resonance card itself, so it is always visible
chrome_1kseLE9Q18

I also standardized the tooltip for conditionals to appear above the icon

### DIFF
--- a/src/Types/sheet.ts
+++ b/src/Types/sheet.ts
@@ -10,7 +10,6 @@ export interface DocumentSection {
     icon?: Displayable | ((data: UIData) => Displayable);
     action?: Displayable;
   }
-  fieldsDescription?: Displayable | ((data: UIData) => Displayable);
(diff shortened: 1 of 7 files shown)


src/Types/sheet.ts
```diff
@@ -10,7 +10,6 @@ export interface DocumentSection {
     icon?: Displayable | ((data: UIData) => Displayable);
     action?: Displayable;
   }
-  fieldsDescription?: Displayable | ((data: UIData) => Displayable);
```
frzyc, line 13: Didn't you add this to add descriptions to the headers of fields? why remove it now?

PR #372, opened 2022-05-10
Partially resolve #338

### LINKED ISSUE #338: Better handling of missing artifacts when generating a build
I'm not sure if the system could even accomodate this-
But currently if you cannot meet the build criteria due to a missing artifact slot the builder fails with the message "Current configuration will not generate any builds. Please change your Artifact configurations, or add/include more Artifacts."

Is it possible to still genenerate a build with 4 or less artifacts since we now have "placeholder" icons that could be used in the nano card display for empty slots. This may prompt users to question "why don't I have a goblet in this build?" and then realise there are no goblets in their database for example.

While this might not be a priority issue - the game doesn't force you to have 5 or even any artifacts equipped on a character.
Having 2x Glad and 3x empty slots still gives you the valid 2-set effect for example.

I think the Optimizer should still generate the "best" build with what it has to work with, even if that's only 4 artifacts equipped to a character.

lantua:
This allows builds that don't have all 5-artifacts upon optimization. However, it still won't generate builds that are impossible, e.g., 4EoSF builds when there are no EoSF artifacts.

### DIFF
--- a/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/background.ts
+++ b/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/background.ts
@@ -128,7 +128,7 @@ function refresh(force: boolean): void {
 type Stats = { [key in MainStatKey | SubstatKey]?: number }
 export type ArtifactBuildData = {
   id: string
-  set: ArtifactSetKey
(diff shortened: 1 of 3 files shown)


lantua:
PS I haven't checked the UI that thoroughly.

PR #418, opened 2022-05-31
fix double "with" in readme

RockRoller01:


frzyc:
man, I was was wondering when someone would complain about this LOL. I put it in here a year ago as a joke. No one reads README sadge
image

frzyc:
Updated in `8.7.3`

Issue #422, opened 2022-06-01
Tooltip crashes on first hover


nguyentvan7:
Using this JSON
https://discord.com/channels/785153694478893126/787340960173391903/981374961170145330
hovering over any tooltip on Xingqiu's page will cause a crash

If you open the formula dropdown, then hover the tooltip, it will not crash anymore

lantua:
This is due to `Translate` component in `KeyMap.get`. Not sure why, but you can't `useTranslation` inside the tooltip if the namespace hasn't been loaded.

frzyc:
yep. here is the fix. 
![image](https://user-images.githubusercontent.com/1754901/171343608-9374f49b-3746-497e-a141-fa2b3ccd3254.png)

However, I'm not sure how or why this works...

frzyc:
Wrapping the tooltip content in a suspense seems to fix it. Im not sure ultimately what is happening underneath that is causing this, but should be a good enough fix for now.

PR #423, opened 2022-06-01
Reenable `pivot` in formula text

lantua:


lantua:
This would also effectively handle (not really fix) #422 since all top-level formulas would now be generic, non-translated text.

PR #425, opened 2022-06-03
Various fixes

### LINKED ISSUE #264: Allow clicking on teammates on the build tab to swap to their build tab
- This is more of a QoL change as it will allow faster optimization of each team member rather then opening the character selector each time.

image

### LINKED ISSUE #270: Even at full width Yae's talents wrap when 'compare' is turned on
image

### LINKED ISSUE #314: Build constraints input box and some conditionals can be changed  during the build process
image

### LINKED ISSUE #404: Feature Request: Add jump to character build page by clicking on character portrait in priority list
My current build process is:
1. Go to character tab
2. Click sort by Name > Descending
3. Select character
4. Go to build tab
5. Generate 
6. Click on "Use Equipped" cog icon to pop up the modal priority list
7. Identify the next character on my priority list
8. Exit the modal window
9. Click the character select button on the build page
10. Click sort by Name > Descending
11. Select next highest priority character
12. Generate
13. Repeat steps 6 through 12 for the next 20 characters

In lieu of having to go through steps 7 through 11, can we click on the character icon in the priority list modal window and have it jump to that characters build page?

This would eliminate at least three clicks per character to be optimized.

### LINKED ISSUE #410: Bug: Artifact piece name briefly indicates "Unknown Piece Name" on first load/reload
image

### LINKED ISSUE #426: Multiselect filters now close the dropdown after every selection
It used to stay open so you could choose several selections without having to re-open each time.

Is this a bug? Or intentional change of it's function?

CPT2206031828-1540x634

nguyentvan7:
Fix #270 - Text wrapping in build comparison
Before:
image
After:
image
Fix #314 - Disable more buttons while building
Fix #264 - Add teammate click target in optimize tab
image
Fix #404 - Add click target to priority list
image
Fix #410 - Change unknown artifact name to `Skeleton`
image
Fix being able to send a weapon to inventory
Fix cosmetic issue with disabled Autocomplete
I removed the dotted line in this screenshot:
image
Fix being unable to clear an Autocomplete with backspace
Remove clear icon when not needed on Autocomplete
Change hoYoVerse to HoYoverse
Fix #426 - Keep MultiAutocomplete open after select


StainAE86:
> Fix being unable to clear an Autocomplete with backspace

Is this what I'm seeing here?
CPT2206031836-370x604

If you backspace the "None" it just reappears again, but when it does it's no longer highlighted so any input after doesn't auto replace the "None"

nguyentvan7:
> > Fix being unable to clear an Autocomplete with backspace
> 
> Is this what I'm seeing here? CPT2206031836-370x604 [ ](https://user-images.githubusercontent.com/101696552/171819523-91e6c636-6847-4c01-be25-42d8191868e9.gif) CPT2206031836-370x604 [ ](https://user-images.githubusercontent.com/101696552/171819523-91e6c636-6847-4c01-be25-42d8191868e9.gif)
> 
> If you backspace the "None" it just reappears again, but when it does it's no longer highlighted so any input after doesn't auto replace the "None"

Yea now the None should not longer reappear automatically

frzyc:
one of my concern with "click on this to go to another page" is that sometimes the element isnt really evident that its a button. so changing mouse, and hover animation needs to be added at least. and also, I don't want to create a "landmine UI", where users are afraid of clicking on anything because it keep leading them somewhere else unintentionally...

nguyentvan7:
Cursor style and hover effects are already handled with how its implemented. I think adding a prompt would be pretty unintuitive. I agree adding click targets indiscriminately would contribute to a landmine UI, but these targets seem to be common workflows for users (optimizing whole teams, or optimizing through the priority list)

StainAE86:
> I also don't want to create a "landmine UI", where users are afraid of clicking on anything because it keep leading them somewhere else unintentionally...

I think this is an intuitive change that works with the flow of optimizing teams and priority.
The curser will change to a pointer to indicate its a link so anyone familiar with how the internet works should understand they can click it.
Plus clicking it by mistake or without knowing what it'll do won't have any irreversible effects or even clear the current page settings so I don't see an issue.

src/Components/FieldDisplay.tsx
```diff
@@ -61,7 +61,10 @@ export function NodeFieldDisplay({ node, oldValue, suffix, component }: { node:
   const formulaTextOverlay = !!node.formula && <QuestionTooltip title={<Typography><Suspense fallback={<Skeleton variant="rectangular" width={300} height={30} />}>{fieldFormulaText}</Suspense></Typography>} />
   return <Box width="100%" sx={{ display: "flex", justifyContent: "space-between", gap: 1 }} component={component} >
     <Typography color={`${node.info.variant}.main`} sx={{ display: "flex", gap: 1, alignItems: "center" }}>{!!isTeamBuff && <Groups />}{icon}{fieldText}{suffix}</Typography>
-    <Typography sx={{ display: "flex", gap: 1, alignItems: "center" }} >{fieldVal}{formulaTextOverlay}</Typography>
+    <Box sx={{ display: "flex", gap: 1, alignItems: "center" }}>
+      <Typography noWrap>{fieldVal}</Typography>
```
frzyc, line 65: noWrap will make it so that extra text gets `...`'d

nguyentvan7: I know the API says that, but I wasn't able to get that to happen even in the most extreme cases (smallest column, with an impossibly large number on Yae's skill)
image

nguyentvan7: Speaking about this particular change, do we need to wrap this in FieldDisplayList or similar, so we can get alternating row colors for clarity?
image

frzyc: i guess we can, at one point that alternating color was conflicting with the background, but now i dont think there is an instance

nguyentvan7: I'll add it in tonight. After that, should be good to go

PR #445, opened 2022-06-15
Don't filter if there is no slotkey selected

theqp:


frzyc:
Will be superseded by #447

PR #453, opened 2022-06-16
Minor optimization

lantua:
just some low-hanging fruits


lantua:
With this, the workers would have up to 8 commands queue up, subjected to availability. Also, if the results of the splitting works are too small, they are consumed directly by the worker, reducing further message passing. There seems to be a slight overhead overall when there are a few exclusions (1-3% AFAICT), but it should help a lot w/ rainbow build generations.

PR #454, opened 2022-06-17
Branch and bound

tooflesswulf:
B&B is a powerful global optimization method, especially for combinatorial optimization tasks like what GO does.

This implementation achieves a respectable speed of 15-30s for zero-filter (no main stat filters, no level filter) problems of typical size on the order of 10,000,000,000 combinations. See the attached video for a 1,000,000,000,000 build problem solved in 13 seconds! For math details, feel free to chat with me on discord.

The bulk of my code, particularly the math portion, is isolated in various files under `src/Formula`. To integrate my math into GO's optimization pipeline I had to mess with the program structure a bit. In particular, we have three primary changes:
1. To facilitate efficient caching of the expensive calculations used for the BNB algorithm's bounding step, I change the main passed object from `RequestFilter` to `SubProblem`.
2. Because the bulk of the work is now the splitting and no longer has anything to do with enumeration, I changed the workers' message passing logic and added a new `SplitWork` message to share the splitting work between different threads.
3. There's a lot of overhead associated with splitting each SubProblem, so to avoid over-splitting any execution branch (), I let the `ComputeWorker` handle the ArtifactSetExclusion logic.

---

The current state of the code: All my math functions are ready to look at. `SplitWorker.ts` should also okay.

`ComputeWorker.ts` is currently a huge mess because I'm trying to verify that all the substitutions & the calculated upper bounds are valid. The set filtering logic hasn't been fully tested but I'm reasonably confident it works.

`TabOptimize/index.tsx` is also a mess, but I think (?) the original logic should mostly be intact.

---

1 trillion builds solved in 13 seconds!!!

https://user-images.githubusercontent.com/3460254/174259664-e9ef5e04-d385-4286-af9b-4e4e5010dbf2.mp4


tooflesswulf:
Closing PR cuz its too old lmao

PR #542, opened 2022-08-05
Update strings

nguyentvan7:
* Update reaction string pipeline to use game-translated strings
* Fix Zhongli constellation name from pipeline
* Standardize reaction-based strings

### DIFF
--- a/pipeline/Data/Characters.ts
+++ b/pipeline/Data/Characters.ts
@@ -60,13 +60,15 @@ const char = {
         thunderDMG: [3560985918, "skillParam"]
       }
     }
-  }
+  },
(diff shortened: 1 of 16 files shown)


frzyc:
Huh, I would think the zhongli fix be a spoiler, no?

nguyentvan7:
Could remove it, but I feel like anyone who:
1. Is using the optimizer
2. Has Zhongli
it would be pretty shocking for that to be a spoiler. Also the wiki just plainly lists the constellation, which I feel like has more of an audience size, especially for casual people who may not be caught up on story quests (but again, this is one of the first story quests in the game)

PR #595, opened 2022-09-01
limit clone_genshindata.js to depth 1

eeeqeee:


nguyentvan7:
The `git pull` might break if there are ever any merge commits iirc. But it's probably reasonable to assume that won't happen?

eeeqeee:
I can change it to fetch + reset instead if that's better

PR #597, opened 2022-09-03
Move mount path and print server url

eeeqeee:
Plain `npm build` doesn't work with the testserver, so this addresses that

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "prebuild": "npm run gen-files",
     "build": "react-scripts build",
     "test": "react-scripts test",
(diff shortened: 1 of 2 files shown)


frzyc (commented):
I think translations got a little weird last time I tried this(months ago) 
https://github.com/frzyc/genshin-optimizer/blob/a96f36d5f0260d72d1f12e82b415a0e1a06f52c9/src/i18n.ts#L62

eeeqeee:
`npm build` uses the line below <https://github.com/frzyc/genshin-optimizer/blob/a96f36d5f0260d72d1f12e82b415a0e1a06f52c9/src/i18n.ts#L65>

I just tested it locally and it seems to work

PR #620, opened 2022-09-14
Correctness 2: noscopy

### LINKED ISSUE #619: Correctness issue when optimal builds are found early
The interim reports aggregation is not idempotent: https://github.com/frzyc/genshin-optimizer/blob/9d5046fc79e06b5ec0640a6125e0f61d31e318cc/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/index.tsx#L204-L205 If multiple messages include the same near-optimal build values, those build values could push others down than it needs to. At the same time, the interim reports generated by `ComputeWorker` may use the identical builds multiple times if those builds are already optimal or near-optimal: https://github.com/frzyc/genshin-optimizer/blob/9d5046fc79e06b5ec0640a6125e0f61d31e318cc/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/ComputeWorker.ts#L98-L102 and thus requires idempotence at the receiver.

This is a correctness issue as it may lead to over-pruning in scenarios where top builds are found early.

lantua:
- Fix #619
- Remove unnecessary build checking

### DIFF
--- a/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/BackgroundWorker.ts
+++ b/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/BackgroundWorker.ts
@@ -13,9 +13,9 @@ onmessage = ({ data }: { data: WorkerCommand }) => {
   switch (command) {
     case "setup":
       id = data.id
-      const callback = (interim: InterimResult) => postMessage({ id, ...interim })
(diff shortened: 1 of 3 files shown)


lantua:
Added `source` field to the interim reports, so that the aggregation can remove old data from that source. This makes the aggregation idempotent w.r.t. that `source`.

PR #627, opened 2022-09-18
Partially Incorporate BnB

lantua:
This partially reimplements the main portion of #454 using most of the functionality already presented in WR, keeping it mostly DRY. The heuristic for the "branching" process is different, but the improvement on that front could come later.


lantua:
Seems to still ned more work

lantua:
Still need some work to handle variable duplicity.

lantua:
Since BNB split worker has a higher overhead than the default split worker, I increased `maxRequestFilterInFlight` so that we can have extra headroom. In addition, I also increased `minFilterCount` so that each `ComputeWorker` can spend more time on the work w/o waiting on the `SplitWorker`.

The parameters may change again as newer split workers are implemented, and the overhead characteristic changes.

frzyc (approved):

PR #648, opened 2022-09-24
constantFold removes constant branches

tooflesswulf:
`( Read[BraveHeart] >= 2 ? 0 : 0 )` becomes `0`


lantua (approved):
No formula will hit this case yet, but many of our future optimizations will. It should be ok to include this early.

PR #655, opened 2022-09-27
Add transformative reactions based on infusion, remove legacy multitargets

### LINKED ISSUE #566: Bennett C6 infused characters don't have Overloaded as a target
Doesn't work using the Infusion hitmode or with C6 Bennett in the team.

image

### LINKED ISSUE #618: Remove Kuki/Venti/Razor combined dmg formulas
Enabling aggravate, vape or melt for these formulas will result in a completely inaccurate calculation, as not each hit of these combined formulas should be able to react. Probably best to remove these with the addition of multi-opt. With sharing multi-opt targets coming back, we can share a multi-target for users to recreate these formulas easily.

Removing these nodes would probably require a DB migration so user multi-targets are not suddenly missing a target or something like that. Probably best to pair with #581, since that also requires a DB migration

nguyentvan7:
## Enhancements
* Resolves #566 - Add transformative reactions based on infusion

## Fixes
* Resolves #618 - Remove Kuki, Venti and Razor full burst formula to prevent bad interactions with global reactions


src/Formula/reaction.ts
```diff
@@ -1,97 +1,114 @@
...
+      `${e}Crystallize`,
+      infoMut(
+        checkCharEleAndInfusion(charElement, ["geo", e], prod(percent(2.5), crystallizeHit)),
+        { key: `${e}_crystallize`, variant: e })
+    ])),
   }
 }
+
+// Check the character's element and infusion against an array of elements. If any matches, the nodes value will be `nodeToReturn`.
+function checkCharEleAndInfusion(charElement: ElementKey, elementsToMatch: ElementKey[], nodeToReturn: NumNode): NumNode {
```
frzyc, line 110: Why are we passing in character's element key here? why wouldnt you be able to access `input.charEle`?

src/Formula/reaction.ts
```diff
@@ -1,97 +1,114 @@
...
-    burgeon: trans.burgeon,
-  },
-  cryo: {
-    superconduct: trans.superconduct,
-    shattered: trans.shattered,
-  },
-  dendro: {
-    burning: trans.burning,
-    bloom: trans.bloom,
+export function getReactions(charElement: ElementKey) {
```
frzyc, line 50: My opinion is that we should make reactions here pretty much static object, instead of a generation function. construct reactions by reading `input.charEle` and `input.infusion` should be enough?

src/Data/Characters/Venti/index.tsx
```diff
@@ -126,12 +126,12 @@ const dmgFormulas = {
     full7: unequal(condBurstAbsorption, undefined, sum(
       prod(dmgNode("atk", datamine.burst.baseDmg, "burst"), 20),
       prod(dmgNode("atk", datamine.burst.absorbDmg, "burst", { hit: { ele: condBurstAbsorption } }), 15),
-      prod(lookup(condBurstAbsorption, objectKeyMap(absorbableEle, ele => reactions.anemo[`${ele}Swirl`]), naught), 7)
+      prod(lookup(condBurstAbsorption, objectKeyMap(absorbableEle, ele => getReactions(elementKey)[`${ele}Swirl`]), naught), 7)
```
frzyc, line 129: Lets just do #618 here and remove the venti formula, since people will try to ducttape this into multi-config for venti...

nguyentvan7:
I moved `infusionNode` to `Formula/index.tsx` and made it all static. Should we drop Lantua's `empty` commit and save for another time?

lantua:
If the new approach can work w/o, let's just drop it

nguyentvan7:
`{"name":"Venti Full Burst","targets":[{"weight":20,"path":["burst","base"],"hitMode":"avgHit","bonusStats":{}},{"weight":15,"path":["burst","absorb"],"hitMode":"avgHit","bonusStats":{}},{"weight":7,"path":["reaction","electroSwirl"],"hitMode":"avgHit","bonusStats":{}}]}`
`{"name":"Kuki Burst 7-tick","targets":[{"weight":7,"path":["burst","singleDmg"],"hitMode":"avgHit","bonusStats":{}}]}`
`{"name":"Kuki Burst 14-tick","targets":[{"weight":14,"path":["burst","singleDmg"],"hitMode":"avgHit","bonusStats":{}}]}`

src/Data/Characters/Razor/index.tsx
```diff
@@ -298,13 +295,13 @@ const sheet: ICharacterSheet = {
       text: trm("fullBurstDMG.description"),
     }, ct.headerTemplate("burst", {
       fields: [{
-        node: infoMut(dmgFormulas.burst.fullBurstDmg1, { key: `char_${key}:fullBurstDMG.dmg1` })
+        node: infoMut(dmgFormulas.burst.fullBurstDmg1, { key: `char_${key}:fullBurstDMG.dmg1`, variant: "invalid" })
```
frzyc, line 298: is there any point in keeping these since we have multi-target?? lets keep the "simple" targets as truly simple targets, or we have too much nesting.

frzyc (approved):

PR #662, opened 2022-09-30
Allow `tally` node with more flexible `accu`

lantua:
Current `tally.*` nodes are assumed to use the `add` accumulator. This may prove inadequate in the near future. That assumption is now lifted, with the tradeoff that any non-standard tally nodes (ones not declared in `index.ts:tally`) must be accessed via `getTally` function.

Adding a tally node to a `Data` is otherwise the same; values are added to `[teamBuff, tally, <name>]` field.


lantua:
It seems `getTally` usage is somewhat confusing as it both registers a new `tally` and fetches an existing one. So I instead remove it and requires that `tally` namespace be wholely declared upfront. Should we need to defer a tally node declaration to after `index.ts` (what happened back in NO4), we can add them back later. In which case we will have a separate `registerTally` and `getTally`.

frzyc (approved):

nguyentvan7 (approved):

PR #667, opened 2022-10-02
fix characterKey on documentation page

frzyc:
- Fix documentation of the `characterKey` to have all variant of the traveler, instead of `Traveler` as intended.

### DIFF
--- a/src/PageDocumentation/index.tsx
+++ b/src/PageDocumentation/index.tsx
@@ -1,6 +1,6 @@
 import { ArrowRightAlt } from "@mui/icons-material";
 import { Box, CardContent, Divider, Grid, Link as MuiLink, Skeleton, styled, Tab, Tabs, Typography } from "@mui/material";
-import React, { Suspense } from "react";
+import { Suspense } from "react";
 import ReactGA from 'react-ga4';
 import { useTranslation } from 'react-i18next';
 import { Link, Route, Routes, useMatch } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


frzyc:
Splitting the Travelers into the travelers also means that the GO exports are not fully GOOD, but since GOOD is only used as a scanner input format, this is relatively benign.

src/PageDocumentation/index.tsx
```diff
@@ -203,7 +203,7 @@ function ArtifactSetKeyPane() {
 }
 function CharacterKeyPane() {
   const { t } = useTranslation([...new Set(allCharacterKeys)].map(k => `char_${k}_gen`))
-  const charKeysCode = `type CharacterKey\n  = ${[...new Set(allCharacterKeys)].sort().map(k => `"${k}" //${t(`char_${k}_gen:name`)}`).join(`\n  | `)}`
+  const charKeysCode = `type CharacterKey\n  = ${[...new Set(locationCharacterKeys)].sort().map(k => `"${k}" //${k === "Traveler" ? "Traveler" : t(`char_${k}_gen:name`)}`).join(`\n  | `)}`
```
nguyentvan7, line 206: Use `charNames_gen.json` so we don't have to load every character's translation json

nguyentvan7, line 206: Also, can use Traveler's translated name? Just have it format as something like
`// Aether/Lumine`

nguyentvan7: While we are at it, we could also update the artifact and weapons to use `artifactNames_gen.json` and `weaponNames_gen.json`

nguyentvan7 (approved):

PR #668, opened 2022-10-02
Allow artifacts in build results to be edited.

### LINKED ISSUE #637: Allow artifacts to be edited directly from the optimization result
Say the optimizer suggests an artifact and I choose to  immediately upgrade that artifact in game. I would like to reflect that upgrade without having to rescan all of my artifacts or go the process of finding the artifact in the GO  inventory screen.

image

frzyc:
Resolve #637 - Allow artifacts in build results to be edited.


nguyentvan7:
Should we disable editing the set? I edited the feather to be archaic petra, and now the optimization still says eosf 4p, and the burst still has the 4p bonus
image

nguyentvan7 (changes requested):
When I edit an artifact, the build stats don't update. Is that intended? (somewhat related to my previous comment I guess?)

frzyc:
yeah, there are a ton of weird interactions here. I need to look at the update chain for some of these stuff, because the build results UI need to register listener for everything that can change

nguyentvan7 (approved):

PR #673, opened 2022-10-05
Retype

lantua:
Some house cleaning regarding the types of functions in `Formula`, fixing some typing bugs along the way.

- Fix a bug that causes `objectMap` to use an incorrect variation
- Update types of util functions
- Remove `AnyNode` and `Base.operands` from type decl

### DIFF
--- a/src/Types/global.d.ts
+++ b/src/Types/global.d.ts
@@ -1,26 +1,17 @@
-
 declare global {
-    type Displayable = string | JSX.Element
-    type Dict<K extends string | number, V> = {
-        [key in K]?: V
-    }
-    type StrictDict<K extends string | number, V> = {
-        [key in K]: V
-    }
+  type Displayable = string | JSX.Element
+  type Dict<K extends string | number, V> = Partial<Record<K, V>>
+  type StrictDict<K extends string | number, V> = Record<K, V>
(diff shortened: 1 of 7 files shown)


src/Types/global.d.ts
```diff
@@ -1,26 +1,17 @@
...
-    type Displayable = string | JSX.Element
-    type Dict<K extends string | number, V> = {
-        [key in K]?: V
-    }
-    type StrictDict<K extends string | number, V> = {
-        [key in K]: V
-    }
+  type Displayable = string | JSX.Element
+  type Dict<K extends string | number, V> = Partial<Record<K, V>>
+  type StrictDict<K extends string | number, V> = Record<K, V>
```
frzyc, line 4: why are we keeping `StrictDict` if `Record` is 4 chars shorter?

lantua: Let's keep them for now. Renaming `StrictDict` covers a lot more ground w.r.t. this PR, which mostly stays within `Formula`.

lantua: If you want to rip the bandage, let's pick the pair we'll use moving forward.

frzyc (approved):

PR #678, opened 2022-10-06
Fix laggy re-rendering in optimization tab

### LINKED ISSUE #677: Changing optimization options triggers update on build results, causing UI lag
Because the builds are stored in `buildSetting`, changing the `buildSetting` causes update to propagate to the build result component, causing needless updates.

frzyc:
Resolve #677 - Move build results to separate database entry.


nguyentvan7:
The UI is still faster when there are no builds vs having 10 builds. Not sure how to pinpoint exactly what is happening

frzyc:
more builds = more UI elements, images, and more `Data` to contend with. This PR isnt intended to address all UI issues here, just the data path that updates the build results unnecessarily.

nguyentvan7 (approved):

PR #665, opened 2022-10-02
Update database handling when switching between different databases

frzyc:
- Allow databases to switch their storage, while maintaining their cached data.
- Database swapping no long have expensive new database initialization steps, since they are swapped "hot".
- Use the data entry classes in database to set/unset data in storage, thus extra data in the localstorage are no longer being bundled in the data of other databases.

PR #683, opened 2022-10-10
Add patch notes link

nguyentvan7:
## Enhancements
* Add a "Patch Notes" link to the header throughout the app, and quick links card on the home page
* Show a red badge indicator to the patch notes link in the header if a newer version is detected, which can be dismissed by hovering over the link
* Add `latestVersion` field to `dbMeta` to track latest version a DB was opened in

### DIFF
--- a/src/Database/DataEntries/DBMetaEntry.ts
+++ b/src/Database/DataEntries/DBMetaEntry.ts
@@ -6,14 +6,16 @@ import { IGO, IGOOD, ImportResult } from "../exim";
 interface IDBMeta {
   name: string,
   lastEdit: number,
-  gender: Gender
+  gender: Gender,
+  latestVersion: string,
 }
 
 function dbMetaInit(): IDBMeta {
   return {
     name: `Database`,
     lastEdit: 0,
-    gender: "F"
+    gender: "F",
+    latestVersion: "1.0.0"
   }
 }
 
@@ -23,19 +25,20 @@ export class DBMetaEntry extends DataEntry<"dbMeta", "dbMeta", IDBMeta, IDBMeta>
   }
   validate(obj: any): IDBMeta | undefined {
     if (typeof obj !== "object") return
-    let { name, lastEdit, gender } = obj
+    let { name, lastEdit, gender, latestVersion } = obj
     if (typeof name !== "string") name = `Database ${this.database.storage.getDBIndex()}`
     if (typeof lastEdit !== "number") console.log("lastEdit INVALID")
     if (typeof lastEdit !== "number") lastEdit = 0
     if (!genderKeys.includes(gender)) gender = "F"
+    if (typeof latestVersion !== "string") latestVersion = "1.0.0"
 
-    return { name, lastEdit, gender } as IDBMeta
+    return { name, lastEdit, gender, latestVersion } as IDBMeta
   }
   importGOOD(go: IGO & IGOOD, result: ImportResult): void {
     const data = go[this.goKey as any]
     if (data) {
-      // Don't copy over lastEdit data
-      const { lastEdit, ...rest } = data
+      // Don't copy over lastEdit or latestVersion data
+      const { lastEdit, latestVersion, ...rest } = data
       this.set(rest)
     }
   }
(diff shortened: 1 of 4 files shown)


nguyentvan7:
Not sure if we should consider just storing the `latestVersion` in `localStorage` but outside of the exported database?

src/Database/DataEntries/DBMetaEntry.ts
```diff
@@ -23,19 +25,20 @@ export class DBMetaEntry extends DataEntry<"dbMeta", "dbMeta", IDBMeta, IDBMeta>
...
-    let { name, lastEdit, gender } = obj
+    let { name, lastEdit, gender, latestVersion } = obj
     if (typeof name !== "string") name = `Database ${this.database.storage.getDBIndex()}`
     if (typeof lastEdit !== "number") console.log("lastEdit INVALID")
     if (typeof lastEdit !== "number") lastEdit = 0
     if (!genderKeys.includes(gender)) gender = "F"
+    if (typeof latestVersion !== "string") latestVersion = "1.0.0"
 
-    return { name, lastEdit, gender } as IDBMeta
+    return { name, lastEdit, gender, latestVersion } as IDBMeta
```
frzyc, line 35: Shouldnt store latestVersion here, since `dbMeta` is per databse

src/Header.tsx
```diff
@@ -112,13 +124,37 @@ function HeaderContent({ anchor }) {
...
+          label={(isLarge || !resize) && t(i18Key)}
+        />
+      )}
       <Box display="flex" alignItems="center">
         <Chip color="success" label={name} />
       </Box>
 
       <Box flexGrow={1} />
-      {links.map(({ i18Key, href, label, icon }) => <Tab key={label} component="a" href={href} target="_blank" icon={icon} iconPosition="start" onClick={e => ReactGA.outboundLink({ label }, () => { })} label={isLarge && t(i18Key)} />)}
+      {links.map(({ i18Key, href, label, icon }) =>
```
frzyc, line 143: Implement the patch link as a separate Component, instead of jerryrigging it into existing button flow(its behaviour is too complex)

Put the patch notes "seen" handling into this component instead of scattering it in several places

src/Header.tsx
```diff
@@ -186,3 +222,16 @@ function MobileHeader({ anchor, currentTab }) {
     <Toolbar id={anchor} />
   </>
 }
+
+function isDbVersionOlder(dbVersion: string) {
```
frzyc, line 226: just store the patch as a string, and compare current patch string vs stored string. no need to amortize into a number, when will we never have a decrease in patch number?

nguyentvan7: When we change minor or major, then patch will reset to 0. If the stored version was 8.9.1, and the new version is 8.10.0, then comparing just patch will not mark the patch notes as new.

Also, if someone last used GO on 8.9.2, then comes back on 8.10.1, it will also not be marked as new.

If I just compare the whole string, then the move from something like 8.9.0 to 8.10.0 will not be marked as new.
image

frzyc: why are you checking if a version number is higher or not?
image

frzyc: Also, storing the data in localstorage is "correct", but it might get destroyed by the database bundling, causing weird behaviour. 
https://github.com/frzyc/genshin-optimizer/pull/665 is suppose to fix that.

src/Header.tsx
frzyc: Also, does it make sense to use up valuable navbar realestate? Don't we already display the version number on the site? why not make that the link

nguyentvan7: I didn't like it in the footer because it's not super visible (have to scroll to the bottom of the site and not many people are there). The header is a bit tight squeezed, but I prefer that than putting it in the header. I could make the patch notes an icon-only (with tooltip) no matter what if that is any better

frzyc: There might be something in the future that will use up that top right space of navbar, so im hesitant to add anything emergent there. I think patchnotes on the home page, and the version number as a link is sufficient. most of the "new" features are easily apparent anyways.

PR #692, opened 2022-10-15
improve type inference for `mapFormulas()`

tooflesswulf:
I got annoyed by VsCode constantly complaining that `n` could be a `StrNode` in the following example.

```ts
// nodes is NumNode[]
mapFormulas(nodes, n => n, n => {
  // stuff with n
})
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "genshin-optimizer",
-  "version": "8.20.3",
+  "version": "8.24.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "genshin-optimizer",
-      "version": "8.20.3",
+      "version": "8.24.0",
       "hasInstallScript": true,
       "dependencies": {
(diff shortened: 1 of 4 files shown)


lantua:
This is still force-casting albeit on a `mapFormula` level. If we're improving on this front, it might be better to add parametric type for the "permissible leaf", e.g., `Base<Leaf>`, `Threshold<ConditionLeaf, ValueLeaf>`. Such change could go directly into `master`.

If the type-checker proves inadequate to do that, we may need to do something like this, though I'm not 100% sure if that is advisable just yet.

Meanwhile, one can still do

```
mapFormula(..., _f => {
  const f = _f as NumNode
})
```

which is not the best, but still keep the casting explicit.

lantua:
Let's explore a bit the parametric "permissible leaf nodes" I mentioned. If it's a dead end, we'll merge this.

lantua:
We ended up using the alternate approach (#698) instead, which should also give a robust type inference.

PR #693, opened 2022-10-15
Fix mistake in skill name of C2 Hu Tao

littlef0x:
C1 Hu Tao:
图片

C2 Hu Tao:
图片

> char_HuTao_gen:constellation2.skillParams.2

I'm not familiar with TypeScript. So please check whether this is the correct way to fix it. Thanks.


frzyc:
Not exactly. Something else needs to be done. https://github.com/frzyc/genshin-optimizer/issues/695

PR #722, opened 2022-10-27
Add labels to graph

### LINKED ISSUE #713: Add Graph Axis labels
Already suggested in #502 however should be bumped to the top of the list.
With the introduction of Graph Opt. Target vs. Opt. Target it's difficult to quickly read the graph and tell which axis is which:

image

Maybe put the name of the character within the Graph as well since its likely people will start sharing their Graph results:

image

nguyentvan7:
## Enhancements
* Resolve #713 - Add labels to comparison graph

Notes: This feels pretty jank, but I guess I already do similar stuff in `ConditionalSelector:getCondName`

### DIFF
--- a/src/Components/EnemyEditor.tsx
+++ b/src/Components/EnemyEditor.tsx
@@ -1,6 +1,7 @@
 import { CheckBox, CheckBoxOutlineBlank, ExpandMore, Replay } from "@mui/icons-material";
 import { Box, Button, CardContent, Chip, Collapse, Grid, Typography } from "@mui/material";
 import { useCallback, useContext, useState } from 'react';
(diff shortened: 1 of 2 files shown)


nguyentvan7:
Just to reiterate, I am doing it in this way because rechart only allows rendering certain components within the chart area. Felt like the best way to do this was to grab the translation string, albeit jankily.

https://gaurav5430.medium.com/exploring-recharts-what-can-be-rendered-inside-a-recharts-container-fe7241a6bc4d

frzyc (changes requested):
image

frzyc (approved):

PR #732, opened 2022-10-28
Fix more misc. issues

### LINKED ISSUE #711: Add Rarity filter to "Add New Weapon" popup
Main weapon page has the Rarity filter, but the Add New popup does not - creating a very long list of weapons.

image

### LINKED ISSUE #719: Enemy editor is not expandable in mobile portrait view
IMG_3164

### LINKED ISSUE #720: Separate Bountiful Core from Bloom DMG
Currently on Nilou (or her team) when the "Golden Chalice's Bounty" conditional is active the Bloom DMG gets the A4 Bonus added to it to convert it to the Bountiful Core damage.

However this causes two issues;

1) The wording "Bountiful Core" does not appear as a Optimization Target which can be confusing to users
2) Nilou and her team have the ability to create normal Bloom seeds / do normal Bloom DMG before Nilou activates her skill and after it expires. This makes creating custom Multi-Targets more challenging

This could possibly spark more conversation on how conditionals are handled within Multi-Targets?
As currently the only way to deal with such situations is to disable the conditional and add the stats manually

### LINKED ISSUE #726: Constellation locked Opt Targets remain selected after constellation is disabled
https://discord.com/channels/785153694478893126/787340960173391903/1035139235562389575

### LINKED ISSUE #727: Move unowned artifact sets to bottom of artifact set config
image

### LINKED ISSUE #730: Partial builds toggle not greyed out while optimizing

nguyentvan7:
## Enhancements
* Resolve #711 - Add rarity filter to new weapon popup
## Fixes
* Resolve #719 - Allow enemy editor to wrap
* Resolve #726 - Don't allow optimizing for empty targets
* Resolve #711 - Add rarity filter to new weapon popup
* Resolve #727 - Move unowned arti sets to bottom of config
* Resolve #730 - Disable partial builds toggle while optimizing
* Fix Xiangling C2 showing when not C2+

### DIFF
--- a/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/OptimizationTargetSelector.tsx
+++ b/src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/OptimizationTargetSelector.tsx
@@ -32,7 +32,8 @@ export default function OptimizationTargetSelector({ optimizationTarget, setTarg
   const node: NodeDisplay | undefined = optimizationTarget && objPathValue(data.getDisplay(), optimizationTarget) as any
 
   const invalidTarget = !optimizationTarget || !displayHeader || !node
-    || (optimizationTarget && data.getDisplay()?.[optimizationTarget[0]][optimizationTarget[1]]?.isEmpty)
+    // Make sure the opt target is valid, if we are not in multi-target
+    || (!ignoreGlobal && optimizationTarget && data.getDisplay()?.[optimizationTarget[0]][optimizationTarget[1]]?.isEmpty)
 
   const prevariant = invalidTarget ? "secondary" : node.info.variant
   const variant = prevariant === "invalid" ? undefined : prevariant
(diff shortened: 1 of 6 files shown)


src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/OptimizationTargetSelector.tsx
```diff
@@ -28,10 +30,13 @@ export default function OptimizationTargetSelector({ optimizationTarget, setTarg
   const { database } = useContext(DatabaseContext)
   const displayHeader = usePromise(() => optimizationTarget && getDisplayHeader(data, optimizationTarget[0], database), [data, optimizationTarget, database])
 
+  if (!defaultText) defaultText = t("targetSelector.selectOptTarget")
+
   const { title, icon, action } = displayHeader ?? {}
   const node: NodeDisplay | undefined = optimizationTarget && objPathValue(data.getDisplay(), optimizationTarget) as any
 
   const invalidTarget = !optimizationTarget || !displayHeader || !node
+    || (optimizationTarget && data.getDisplay()?.[optimizationTarget[0]][optimizationTarget[1]]?.isEmpty)
```
frzyc, line 39: For multi-target, there is a case where a user might get a "invalid" target, say for crescent pike, from share/import config. This invalid value should remain in the list.

frzyc:
This should Resolve #733, right?

nguyentvan7:
> This should Resolve #733, right?

I'm moving the fix to #733 to a different PR since there are a few changes to go with that

frzyc (approved):

PR #737, opened 2022-10-30
Optimize the flow of the Optimize tab

### LINKED ISSUE #733: Make Opt Target and Graph Target more obvious
Now that we can graph against any Opt Target it's can be confusing because the eye is drawn to the same UI element as we're been used to using for "Optization Target", but it's now much closer to the "Generate Builds" button.

Yes this is a reading issue - but not all users are native english speakers and many suffer from 'skimming' over text anyway.
image

I suggest grouping the Opt Target and Graph Target together to make it significantly more obvious:
image

nguyentvan7:
## Fixes
* Resolve #733 
* Update wording for Graph Target selector
* Move generate button to the right side, move opt target selector to the left side
* Make the generate button turn into a cancel button; removed the dedicated cancel button
* Disabled Graph Target selector when no opt target has been selected
* Add tooltips for Generate button and Graph Target selector when no opt target has been selected

### DIFF
--- a/public/locales/en/page_character_optimize.json
+++ b/public/locales/en/page_character_optimize.json
@@ -89,5 +89,6 @@
   "targetSelector": {
     "selectOptTarget": "Select an Optimization Target",
     "selectGraphTarget": "Select a Graph Target"
-  }
+  },
+  "selectTargetFirst": "Select an Optimization Target first"
 }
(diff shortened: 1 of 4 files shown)


frzyc:
Make the whole thing a single button group
image

frzyc (approved):

PR #738, opened 2022-10-31
Move `info` to inner nodes for branching utils

### LINKED ISSUE #731: Aggravate is coloring normal attack damage, even when not applicable

lantua:
Branching nodes with an empty branch (so excluding `threshold`, `compareEq`, and `matchFull`) have better ergonomic if we attach `info` only to the non-empty branches. This shouldn't affect existing usages.

Also, Fix #731

### DIFF
--- a/src/Formula/utils.ts
+++ b/src/Formula/utils.ts
@@ -31,6 +31,7 @@ export function percent(value: number, info?: Info): ConstantNode<number> {
 export function infoMut(node: OptNode, info: Info): OptNode
 export function infoMut(node: NumNode, info: Info): NumNode
 export function infoMut(node: StrNode, info: Info): StrNode
+export function infoMut(node: AnyNode, info: Info): AnyNode
 export function infoMut(node: AnyNode, info: Info): AnyNode {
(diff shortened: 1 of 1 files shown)


nguyentvan7 (approved):
Limited testing looked good to me

src/Formula/utils.ts
```diff
@@ -31,6 +31,7 @@ export function percent(value: number, info?: Info): ConstantNode<number> {
 export function infoMut(node: OptNode, info: Info): OptNode
 export function infoMut(node: NumNode, info: Info): NumNode
 export function infoMut(node: StrNode, info: Info): StrNode
+export function infoMut(node: AnyNode, info: Info): AnyNode
```
nguyentvan7, line 34: Oops, one question. Why is this duplicated?

lantua: The last one specifies the type for the function body. The remaining ones (including the duplicate) specify the possible type combinations. Without the duplicate, `infoMut` won't accept the `AnyNode` variant.

frzyc (approved):

Issue #752, opened 2022-11-03
Artifact Set Config is sorting strangely


nguyentvan7:
<img width="572" alt="image" src="https://user-images.githubusercontent.com/36019388/199641598-63b7d560-54c1-450c-9509-f9e77b0a8158.png">

PR #754, opened 2022-11-03
Update artifact set order in optimize page

lantua:
This PR reorders the artifact sets in Artifact Set Configuration (Optimize page). The ordering is as follows, from high to low priorities:

- "Disabled" entries to the back
- Rarity in descending order
- Artifact Sheets order (alphabetical in ascending order)


lantua:
Also resolve #752

PR #755, opened 2022-11-03
Remove pipeline

frzyc:


nguyentvan7:
Saw you getting rid of pipeline imports. What about these ones in all character `index.tsx`

image

frzyc:
> Saw you getting rid of pipeline imports. What about these ones in all character `index.tsx`
> 
> image

This one is intended to know the shape of characterData. The imports i removed should be using the in-repo `const` definition that was wrongly imported.

lantua:
Should we use `data` instead of `dm` in index files? It'll become obscure name afterward otherwise.

frzyc:
> Should we use `data` instead of `dm` in index files? It'll become obscure name afterward otherwise.

I was going to use `data` but there were variables named data in those files, we can always rename later.

nguyentvan7:
`talentValues`, `tv`, etc also work, since `data` variable name is already used

lantua (approved):

nguyentvan7 (approved):

PR #766, opened 2022-11-08
Move resonance to teambuff

### LINKED ISSUE #765: Resonance conditionals do not apply to the whole team
For example, Nahida A1 as a teammate will not be buffed when enabling Dendro resonance conditional EM

nguyentvan7:
## Fixes
* Resolve #765 - Fix resonance conditionals not applying to teammates
* Resonance buff now shows in the Team Buffs card

Something to note, we could do something similar to this for NO4 and other non-stackable buffs I think?

### DIFF
--- a/src/Data/Characters/Nahida/index.tsx
+++ b/src/Data/Characters/Nahida/index.tsx
@@ -211,9 +211,6 @@ const dmgFormulas = {
       triKarmaAddl
     )
   },
-  passive1: {
-    a1InBurst_eleMas
-  },
   passive2: {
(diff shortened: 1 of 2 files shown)


src/Data/Characters/Nahida/index.tsx
```diff
@@ -211,6 +211,9 @@ const dmgFormulas = {
       triKarmaAddl
     )
   },
+  passive1: {
+    a1InBurst_eleMas
+  },
```
frzyc, line 216: Is Nahida change suppose to be here?

Issue #763, opened 2022-11-08
Optimization results become worse when Assuming Level 20 Mainstats


nguyentvan7:
What is extra strange is that the best build (when level assumption is off) is a fully leveled +20 set. Once you set the assumption, the build it outputs is worse. Json at the bottom

**Level Assumption OFF**
<img width="786" alt="image" src="https://user-images.githubusercontent.com/36019388/200469801-0a4c919e-da64-4ba5-9c37-64b70220d127.png">

**Level Assumption at 20**
<img width="785" alt="image" src="https://user-images.githubusercontent.com/36019388/200469687-fd383e57-b27b-42a8-b9f9-8cfb4845f332.png">

[leveled_2022-11-08_03-37-08.txt](https://github.com/frzyc/genshin-optimizer/files/9957388/leveled_2022-11-08_03-37-08.txt)

nguyentvan7:
Narrowed it down. Exclude every artifact except:
* XL's currently equipped artifacts
* Bennett's flower
* XQ's feather
* Raiden's sands
* And the below flower:
<img width="340" alt="image" src="https://user-images.githubusercontent.com/36019388/200478468-140b035f-2aca-4347-b38d-65975963988e.png">

If you optimize for Level 20 assumption, you will get the bad build. If you include every single artifact, but exclude this flower, then builds are correct

StainAE86:
Is the fix here just to ignore "impossible" artifacts? Please tell me this is the fix here.

nguyentvan7:
[min.txt](https://github.com/frzyc/genshin-optimizer/files/9957935/min.txt)

Here is the absolute minimum DB needed to replicate the bug. If you exclude the unequipped flower, then builds will be ok. if you include the unequipped flower, then builds will be broken

StainAE86:
The problem appears have something to do with having 3 line artifacts with multiple rolls of some/each of the substats.

example: even though the substat value isnt being flagged is invalid - it's an impossible situation (either +0 3 line artifact with more than 3 rolls or a +20 with only 3 lines, etc)
![image](https://user-images.githubusercontent.com/101696552/200485858-bd55949f-012b-436d-9984-be41491a15ba.png)

PR #769, opened 2022-11-10
fix cornercase with pruning

frzyc:
Artifact comparasion did not by default make all substats 0, which can result in artifacts being pruned when they are not strictly inferior. This change address this cornercase.


lantua (approved):

lantua:
> I assume this is meant to fix #763? (subtle hint to attach the issue)

yes, that should fix it

PR #774, opened 2022-11-18
Allow toggling min filter, allow min filter on any node

nguyentvan7:
## Enhancements
* Allow setting a minimum build constraint on any display node (also includes a DB migration)
* Allow quickly enabling/disabling min filters

## Fixes
* Fix text alignment/spacing in optimization target selector


nguyentvan7:
If this comes out in the same patch as Layla, just need to update the comments for the migration versions

nguyentvan7:
Marking as draft until I have time to fix this
https://discord.com/channels/785153694478893126/1044854189714591804/1045275888121548840

nguyentvan7:
Ready again

frzyc (approved):

PR #783, opened 2022-11-22
Reload page if worker fails to load

eeeqeee:
This should help when updating gh-pages to a new version
Showing an alert might be better but requires UI work


src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/index.tsx
```diff
@@ -185,6 +185,10 @@ export default function TabBuild() {
     const finalizedList: Promise<FinalizeResult>[] = []
     for (let i = 0; i < maxWorkers; i++) {
       const worker = new Worker(new URL('./BackgroundWorker.ts', import.meta.url))
+      worker.addEventListener("error", e => {
```
frzyc, line 188: In theory, replicable worker errors will always refresh the page, making it hard to trace. 
should probably lock this behind `process.env.NODE_ENV === "production"`.

frzyc:
I agree, this can be a bad stopgap, forcing persistent refreshing. 
I think a alert that user confirm would be the easiest solution.

frzyc (approved):

PR #804, opened 2022-12-06
Stop using undeclared deps

eeeqeee:
`@mui/system` -> `@mui/material`
`react-router` -> `react-router-dom`


src/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -1,6 +1,6 @@
 import { Checkroom, ChevronRight } from '@mui/icons-material';
 import { Button, CardContent, Grid, Skeleton, Typography } from '@mui/material';
```
nguyentvan7, line 2: Can you import `Box` here instead of on its own line as well, ty :) Same goes for other files

eeeqeee: fixed

nguyentvan7 (approved):

PR #839, opened 2022-12-21
Fix partial build equipping not removing artifacts

### LINKED ISSUE #791: Equipping a partial build will not unequip artifacts
Ganyu build before equipping the below build
image
image
Ganyu build after equipping (the goblet should have been removed, but was not)
image

nguyentvan7:
Resolve #791


frzyc:
@nguyentvan7 's original change for posterity 
image

frzyc (approved):

PR #858, opened 2023-01-01
Add build number to graph tooltip

### LINKED ISSUE #855: add build numbering on the optimization target graph
when a point on the optimization target graph is clicked on, if this build is in the build list, its ranking in the build list should be shown. this way, distinguishing between builds would become easier.

nguyentvan7:
* Resolve #855
* Close tooltip after adding a build to the list
* Simplify logic to compare builds in the graph


frzyc:
Why is it a good idea to close the tooltip after adding? wouldnt it be more responsive, that once you add the build, the tooltip updates with the graph#?

nguyentvan7:
I couldn't get the tooltip to reload with the graph # for some reason. some of my changes even further made it worse because the button would also not disable itself after clicking it. Can test that out by just removing the call to close it

frzyc:
Because of the build issue, its not very easy for us to dev this feature locally, I guess its good enough for now, we can readdress this later.

frzyc (approved):

PR #865, opened 2023-01-03
Changes a ton of Strings (mainly Conditionals) for consistency and description accuracy

PolomoPT:
Most of the changes here are for cases where a Conditional was worded as to imply the effect granted was only on the attack causing it (e.g. "On hit") while the buff actually happened after attacking. Therefore, those instances were changed to "After [hit]". Changes for readability of the conditional were also done. Unused strings were deleted. Duplicate strings where there was clearly a principal, used, key, were deleted and their use-cases changed (Prototype Rancour and Liyue Series of weapons).
I also added a string to Raiden called "burstDmg". It should be used to describe the Musou no Hitotachi's DMG instead of the current one, that erroneously uses "Base" (the in-game multiplier won't take into consideration Resolve, that's why it adds "Base". GO does, so it isn't at all base). I couldn't change that myself, so I'm posting on Discord #bugreports this and other changes that should be done.

Please and thank you, Polomo! <3


PolomoPT:
Realized this might be #841 pt. 2 Electric Boogaloo lmao. Van went ahead and fixed part of the issue strings, but there was a whole lot left still

PolomoPT:
These past few were changing something I overlooked while initially changing string tenses. "enemy was hit" is now "enemy has been hit" for readability

nguyentvan7 (approved):

frzyc (approved):

PR #889, opened 2023-01-17
Updating Flower Of Paradise Lost Stacking precent

IceNWinter:
The precent was set to 10% when it should be 25% per stack


nguyentvan7:
I verified in game that it's actually 25% of the base effect. So 10% is correct in this case

PR #905, opened 2023-01-23
Remove use of any for translation component, fix other warnings

nguyentvan7:


apps/frontend/src/app/Types/global.d.ts
```diff
@@ -1,17 +1,18 @@
...
-  }
-}
-
-export { }
+declare global {
+  type Displayable = string | JSX.Element
+  type Dict<K extends string | number, V> = Partial<Record<K, V>>
+  type StrictDict<K extends string | number, V> = Record<K, V>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  type TransObject = any
```
frzyc, line 6: basically sweeps the `any` under the fridge. I guess if we haven't gotten rid of it, this is not a bad solution

frzyc (approved):

PR #906, opened 2023-01-23
Move assets to modules

frzyc:


frzyc:
I also copied the `consts` we use from `apps\frontend\src\app\Types\consts.ts` to a new module. I have not updated the imports in `frontend` because it would a massive amount of code churn

nguyentvan7 (approved):

PR #916, opened 2023-01-27
Remove fontawesome, update header/homepage links

frzyc:
- Convert all the faicons to SVGIcons, convert all svg data from 14x14 to 24x24
- Remove all fontawesome reference/depdencies
- Update homepage quick links
- Update header links
- Minor updates to dependencies

image


apps/frontend/src/app/Components/AdditiveReactionModeText.tsx
```diff
@@ -1,9 +1,12 @@
 import { Box } from "@mui/material";
 import { useTranslation } from "react-i18next";
+import { ElementIcon } from "../KeyMap/StatIcon";
+import { iconInlineProps } from "../SVGIcons";
```
nguyentvan7, line 4: Should we just have `iconInlineProps` be the default props for the icons and then overwrite them as needed?

apps/frontend/src/app/Components/AmpReactionModeText.tsx
```diff
@@ -17,16 +17,16 @@ export const ampReactionMap = {
   }
 } as const
 const sqBadgeStyle = { mx: 0.25, px: 0.25 }
-export default function AmpReactionModeText({ reaction, trigger }: { reaction: AmpReactionKey, trigger?: Variant }) {
+export default function AmpReactionModeText({ reaction, trigger }: { reaction: AmpReactionKey, trigger?: "cryo" | "pyro" | "hydro" }) {
```
nguyentvan7, line 20: can you make a type, instead of repeating it multiple times? Looks like that type would be used in other places too

`type AmpReactionTrigger = "cryo" | "pyro" | "hydro"

apps/frontend/src/app/Components/Artifact/ArtifactFilterDisplay.tsx
```diff
@@ -1,12 +1,13 @@
-import { faBan, faChartLine } from "@fortawesome/free-solid-svg-icons";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { allArtifactSets, allSlotKeys } from "@genshin-optimizer/consts";
 import { BusinessCenter, Lock, LockOpen, PersonSearch } from '@mui/icons-material';
+import BlockIcon from '@mui/icons-material/Block';
```
nguyentvan7, line 3: Is there a reason this isn't just importing `Block` from `@mui/icons-material`?

apps/frontend/src/app/Components/Artifact/ArtifactFilterDisplay.tsx
```diff
@@ -1,12 +1,13 @@
-import { faBan, faChartLine } from "@fortawesome/free-solid-svg-icons";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { allArtifactSets, allSlotKeys } from "@genshin-optimizer/consts";
 import { BusinessCenter, Lock, LockOpen, PersonSearch } from '@mui/icons-material';
+import BlockIcon from '@mui/icons-material/Block';
+import ShowChartIcon from '@mui/icons-material/ShowChart';
```
nguyentvan7, line 4: See above comment

apps/frontend/src/app/Components/Character/CharacterCardPico.tsx
```diff
@@ -1,29 +1,35 @@
...
   const onClickHandler = useCallback(() => characterKey && onClick?.(characterKey), [characterKey, onClick])
   const actionWrapperFunc = useCallback(
     children => <CardActionArea onClick={onClickHandler}>{children}</CardActionArea>,
     [onClickHandler])
   if (teammateSheet && character) {
     const title = <Suspense fallback={<Skeleton variant="text" width={100} />}>
-      <Typography>{teammateSheet.elementKey && StatIcon[teammateSheet.elementKey]} {teammateSheet.name}</Typography>
+      <Typography>
+        <ColorText color={teammateSheet.elementKey}>
+          {teammateSheet.elementKey && <ElementIcon ele={teammateSheet.elementKey} iconProps={iconInlineProps} />} {teammateSheet.name}
```
nguyentvan7, line 30: I'm not a huge fan of coloring the name of the teammate. Element icon is fine though. But up to you

apps/frontend/src/app/Components/StarDisplay.tsx
```diff
@@ -1,10 +1,7 @@
...
-import ColorText from './ColoredText'
-
-export const StarIcon = () => <FontAwesomeIcon icon={faStar} />
+import { Rarity } from '@genshin-optimizer/consts';
+import StarRoundedIcon from '@mui/icons-material/StarRounded';
+import ColorText from './ColoredText';
 export const StarsDisplay = ({ stars, colored = false }: { stars: Rarity, colored?: boolean }) =>
   <ColorText color={colored ? "warning" : undefined} >
-    {stars ? [...Array(stars).keys()].map((_, i) => <StarIcon key={i} />) : null}
+    {stars ? [...Array(stars).keys()].map((_, i) => <StarRoundedIcon key={i} fontSize="inherit" sx={{ verticalAlign: "-10%" }} />) : null}
```
nguyentvan7, line 6: These are a bit smaller than they used to be, could we increase the size?
here: image

prod: image

apps/frontend/src/app/KeyMap/StatIcon.tsx
```diff
@@ -0,0 +1,83 @@
...
+      return <HealingBonusIcon  {...iconProps} />
+    case "cdRed_":
+      return <CdRedIcon  {...iconProps} />
+    case "shield_":
+      return <ShieldStrIcon  {...iconProps} />
+    case "stamina":
+      return <StaminaIcon  {...iconProps} />
+  }
+  for (const e of allElementsWithPhy)
+    if (statKey.includes(e))
```
nguyentvan7, line 58: If you want more switch-case stuff, we can just slice from first char to first `_`, and that should give you the element (if it exists) to switch-case off of

apps/frontend/src/app/KeyMap/StatIcon.tsx
```diff
@@ -0,0 +1,83 @@
+import { allElementsWithPhy, ElementKeyWithPhy } from "@genshin-optimizer/consts"
```
nguyentvan7, line 1: Is there a reason we should put this in KeyMap and not Components, where it used to be?

apps/frontend/src/app/PageArtifact/ArtifactCard.tsx
```diff
@@ -100,18 +104,18 @@ export default function ArtifactCard({ artifactId, artifactObj, onClick, onDelet
...
             </Box>
-            <Typography color="text.secondary" variant="body2">
-              <SlotNameWithIcon slotKey={slotKey} />
+            <Typography color="text.secondary" variant="body2" sx={{ display: "flex", gap: 0.5, alignItems: "center" }}>
+              <SlotIcon iconProps={{ fontSize: "inherit" }} slotKey={slotKey} />
+              {t<string>(`slotName.${slotKey}`)}
             </Typography>
             <Typography variant="h6" sx={{ display: "flex", alignItems: "center", gap: 1 }}>
-              {mainIcon}
+              <ColorText color={ele}><StatIcon statKey={mainStatKey} /></ColorText>
```
nguyentvan7, line 118: Center aligning seems to not be working, looks like these are bottom aligned
image

apps/frontend/src/app/KeyMap/StatIcon.tsx
```diff
@@ -0,0 +1,83 @@
...
+      return <StaminaIcon  {...iconProps} />
+  }
+  for (const e of allElementsWithPhy)
+    if (statKey.includes(e))
+      return <ElementIcon ele={e} iconProps={iconProps} />
+
+  return null
+}
+
+export function ElementIcon({ ele, iconProps = {} }: { ele: ElementKeyWithPhy, iconProps?: SvgIconProps }) {
```
nguyentvan7, line 64: Should we have a `colored` prop to automatically apply the appropriate element color? And also have that same prop in the icon themselves. Then we don't need to wrap it in `ColorText` ourselves everytime we need color

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/ArtifactSetConfig.tsx
```diff
@@ -94,10 +95,10 @@ export default function ArtifactSetConfig({ disabled }: { disabled?: boolean, })
...
-          <Typography>{t`artSetConfig.2rainbow`} <SqBadge color={allowRainbow2 ? "success" : "secondary"}><FontAwesomeIcon icon={allowRainbow2 ? faChartLine : faBan} className="fa-fw" /> {allowRainbow2 ? t<string>("artSetConfig.allowed") : "Excluded"}</SqBadge></Typography>
-          <Typography>{t`artSetConfig.4rainbow`} <SqBadge color={allowRainbow4 ? "success" : "secondary"}><FontAwesomeIcon icon={allowRainbow4 ? faChartLine : faBan} className="fa-fw" /> {allowRainbow4 ? t<string>("artSetConfig.allowed") : "Excluded"}</SqBadge></Typography>
+          <Typography>{t`sheet:2set`} <SqBadge sx={{ display: "inline-flex", gap: 0.5 }} color="success">{allow2} <ShowChartIcon {...iconInlineProps} /> {t<string>("artSetConfig.allowed")}</SqBadge>{!!exclude2 && " / "}{!!exclude2 && <SqBadge sx={{ display: "inline-flex", gap: 0.5 }} color="secondary">{exclude2} <BlockIcon {...iconInlineProps} /> {t<string>("artSetConfig.excluded")}</SqBadge>}</Typography>
```
nguyentvan7, line 98: Can store the SqBadge sx in a const for easy reuse/future modification

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/MainStatSelectionCard.tsx
```diff
@@ -90,9 +95,10 @@ export default function MainStatSelectionCard({ disabled = false, filteredArtIds
                 : "secondary"
               return <Grid item key={mainStatKey} flexGrow={1} xs={((i < 3 && slotKey !== "goblet") || slotKey === "goblet") ? 4 : undefined} >
                 <BootstrapTooltip placement="top" title={<Typography><strong><StatColoredWithUnit statKey={mainStatKey} /></strong></Typography>} disableInteractive>
-                  <Button fullWidth size="small" color={color} sx={{ height: "100%", pointerEvents: disabled ? "none" : undefined, cursor: disabled ? "none" : undefined, display: "flex", gap: 0.5 }}
+                  <Button fullWidth size="small" color={color} sx={{ height: "100%", pointerEvents: disabled ? "none" : undefined, cursor: disabled ? "none" : undefined }}
+                    startIcon={<StatIcon statKey={mainStatKey} />}
```
nguyentvan7, line 99: Great change, makes the icon much more visible

apps/frontend/src/app/SVGIcons/Element/CryoIcon.tsx
```diff
@@ -0,0 +1,4 @@
...
...312 13.78125 13.160156 C 12.90625 13.640625 12.371094 14.570312 12.394531 15.566406 C 12.421875 16.566406 13.003906 17.46875 13.902344 17.902344 C 14.304688 18.074219 14.695312 18.257812 15.085938 17.929688 C 15.464844 17.640625 15.769531 17.648438 16.316406 17.90625 C 14.425781 19.523438 12.949219 21.570312 12.011719 23.875 C 11.058594 21.574219 9.578125 19.535156 7.6875 17.914062 C 8.054688 17.636719 8.5625 17.636719 8.929688 17.914062 C 9.960938 18.507812 11.40625 17.230469 11.597656 16.199219 C 12.019531 14.179688 9.992188 12.589844 8.199219 12.207031 C 7.320312 12 6.421875 11.894531 5.542969 11.691406 C 1.484375 10.863281 0.875 6.753906 2.859375 3.609375 C 4.007812 12.007812 9.71875 9.703125 12 14.339844 C 14.265625 9.761719 20.019531 11.972656 21.164062 3.609375 C 22.554688 6.035156 22.972656 9.644531 20.042969 11.074219 Z M 12.976562 19.972656 L 12 19.320312 L 11.019531 19.980469 L 12.015625 21.648438 Z M 12 17.378906 C 11.6875 18.128906 11.0625 18.699219 10.285156 18.9375 C 10.894531 19.140625 11.5625 18.957031 11.984375 18.472656 C 12.714844 19.035156 13.347656 19.1875 13.777344 18.910156 C 12.980469 18.699219 12.324219 18.136719 12 17.378906 Z M 12 17.378906 " />, "Cryo")
```
nguyentvan7, line 3: This is the Anemo Icon, somehow only this one is messed up

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOverview/EquipmentSection.tsx
```diff
@@ -88,9 +89,10 @@ function ArtSwapCard({ slotKey }: { slotKey: SlotKey }) {
   const { character: { key: characterKey } } = useContext(CharacterContext)
   const { database } = useContext(DatabaseContext)
   const [show, onOpen, onClose] = useBoolState()
+  const { t } = useTranslation("artifact")
   return <CardLight sx={{ height: "100%", width: "100%", minHeight: 300, display: "flex", flexDirection: "column" }}>
     <CardContent>
-      <Typography><SlotNameWithIcon slotKey={slotKey} /></Typography>
+      <Typography sx={{ display: "flex", gap: 1 }}><SlotIcon iconProps={iconInlineProps} slotKey={slotKey} /> {t<string>(`slotName.${slotKey}`)}</Typography>
```
nguyentvan7, line 95: Also not center aligned 
image

apps/frontend/src/app/SVGIcons/Element/AnemoIcon.tsx
```diff
@@ -0,0 +1,4 @@
...
...12 13.78125 13.160156 C 12.90625 13.640625 12.371094 14.570312 12.394531 15.566406 C 12.421875 16.566406 13.003906 17.46875 13.902344 17.902344 C 14.304688 18.074219 14.695312 18.257812 15.085938 17.929688 C 15.464844 17.640625 15.769531 17.648438 16.316406 17.90625 C 14.425781 19.523438 12.949219 21.570312 12.011719 23.875 C 11.058594 21.574219 9.578125 19.535156 7.6875 17.914062 C 8.054688 17.636719 8.5625 17.636719 8.929688 17.914062 C 9.960938 18.507812 11.40625 17.230469 11.597656 16.199219 C 12.019531 14.179688 9.992188 12.589844 8.199219 12.207031 C 7.320312 12 6.421875 11.894531 5.542969 11.691406 C 1.484375 10.863281 0.875 6.753906 2.859375 3.609375 C 4.007812 12.007812 9.71875 9.703125 12 14.339844 C 14.265625 9.761719 20.019531 11.972656 21.164062 3.609375 C 22.554688 6.035156 22.972656 9.644531 20.042969 11.074219 Z M 12.976562 19.972656 L 12 19.320312 L 11.019531 19.980469 L 12.015625 21.648438 Z M 12 17.378906 C 11.6875 18.128906 11.0625 18.699219 10.285156 18.9375 C 10.894531 19.140625 11.5625 18.957031 11.984375 18.472656 C 12.714844 19.035156 13.347656 19.1875 13.777344 18.910156 C 12.980469 18.699219 12.324219 18.136719 12 17.378906 Z M 12 17.378906 " />, "Anemo")
```
nguyentvan7, line 3: See my comment in `ElementIcon` for adding a `colored` prop

nguyentvan7 (changes requested):

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOverview/EquipmentSection.tsx
nguyentvan7, line 95: Also why is there a random `<string>` here? removing it didn't seem to hurt anything

frzyc: um, it explicitly casts the return of `t` as a `string`, which doesnt do much now since we disabled it from returning `null`

apps/frontend/src/app/Components/AdditiveReactionModeText.tsx
frzyc: I'm not sure yet, I've changed this several times when i was refactoring, and either cases doesnt feel right. I think i'll keep this as a shortcut for now

apps/frontend/src/app/Components/Artifact/ArtifactFilterDisplay.tsx
frzyc: Its how the mui Icon documentation lets you copy when you look up icons. so I was too lazy to change their convention, which kind of conflict with our existing import convention. I'd say importing `...Icon` is probably better since it explicitly states that this var is an icon.

apps/frontend/src/app/Components/Character/CharacterCardPico.tsx
frzyc: I'll color the element icon

apps/frontend/src/app/Components/AdditiveReactionModeText.tsx
nguyentvan7: Should we just have an `inline` bool prop? If all the icons use a base `icon` element with that prop, sounds like the best case

apps/frontend/src/app/KeyMap/StatIcon.tsx
frzyc: It will eventually be powered by the keymap lookup in wr2,

apps/frontend/src/app/KeyMap/StatIcon.tsx
frzyc: We can use the ```sx={{color:`${color}.main`}}``` as a slightly better shortcut, still kinda unwieldy

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/ArtifactSetConfig.tsx
frzyc: those can all be removed

apps/frontend/src/app/Components/AdditiveReactionModeText.tsx
frzyc: the issue is the baseline requires changes to `sx` which makes props harder to merge, and prone to overwrite. I'll probably refactor this system later...

apps/frontend/src/app/Components/AmpReactionModeText.tsx
frzyc: Probably doesnt occur often enough to matter. we have too many weird types around.

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOverview/EquipmentSection.tsx
frzyc: I've removed all occurences.

apps/frontend/src/app/Components/AmpReactionModeText.tsx
nguyentvan7: if we have to change this later, i blame you :)

apps/frontend/src/app/Components/Artifact/ArtifactFilterDisplay.tsx
nguyentvan7: So at some point we should move all the icons-material imports to be like this? From what I can tell, bundle size is unaffected by this so its purely for us as devs. It would make the import sections a bit longer, but I guess its kinda whatever for it to be longer

apps/frontend/src/app/KeyMap/StatIcon.tsx
nguyentvan7: Is it difficult to add the `colored` boolean prop? it seems more intuitive, and easier to type

nguyentvan7 (changes requested):
See replies to existing comments

apps/frontend/src/app/Components/Artifact/ArtifactFilterDisplay.tsx
frzyc: perhaps, but I think i have enough churn in this PR

apps/frontend/src/app/KeyMap/StatIcon.tsx
frzyc: yeah. I think eventually we might, but I think this PR already address most of the usage, and there are only like 2-3 occurrences, so I dont see the point of this refactor until its more of a anti-pattern

apps/frontend/src/app/Components/AdditiveReactionModeText.tsx
nguyentvan7: Meh fine I suppose

nguyentvan7 (approved):

PR #922, opened 2023-01-29
fix #918

### LINKED ISSUE #918: Add note to "Multi-Optimization Target Config" popup hinting to where people may find examples
"Note: Community created custom Multi-Optimization Targets can be found within the GO or KQM Discords, however the validity of such configurations cannot be guaranteed. It is the responsibility of the user to confirm the accuracy for their own use case."

frzyc:
Just separates the button from the tooltip, should also work better on mobile as well.
image


frzyc:
Linked to the wrong issue.

PR #923, opened 2023-01-29
Fix "assume main stat level" dropdown + tooltip interaction

frzyc:
Just separates the button from the tooltip, should also work better on mobile as well.
image


nguyentvan7 (approved):
Weird when I tried this it didn't work, but now it does. Oh well

PR #924, opened 2023-01-29
validate character talent level

frzyc:
There are some scanner issues where character talent levels can have invalid values, which can cause funky behaviour.

### DIFF
--- a/apps/frontend/src/app/Database/DataManagers/CharacterData.ts
+++ b/apps/frontend/src/app/Database/DataManagers/CharacterData.ts
@@ -4,7 +4,7 @@ import { validateLevelAsc } from "../../Data/LevelData";
 import { validateCustomMultiTarget } from "../../PageCharacter/CustomMultiTarget";
 import { CustomMultiTarget, ICachedCharacter, ICharacter } from "../../Types/character";
 import { allAdditiveReactions, allAmpReactions, allHitModes, allInfusionAuraElements, charKeyToLocCharKey, InfusionAuraElements, LocationCharacterKey } from "../../Types/consts";
-import { deepClone, objectKeyMap } from "../../Util/Util";
+import { clamp, deepClone, objectKeyMap } from "../../Util/Util";
 import { defaultInitialWeapon } from "../../Util/WeaponUtil";
 import { ArtCharDatabase } from "../Database";
 import { DataManager, TriggerString } from "../DataManager";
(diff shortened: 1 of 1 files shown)


apps/frontend/src/app/Database/DataManagers/CharacterData.ts
```diff
@@ -34,17 +34,17 @@ export class CharacterDataManager extends DataManager<CharacterKey, "characters"
...
 
     if (typeof talent !== "object") talent = { auto: 1, skill: 1, burst: 1 }
     else {
       let { auto, skill, burst } = talent
-      if (typeof auto !== "number" || auto < 1 || auto > 15) auto = 1
-      if (typeof skill !== "number" || skill < 1 || skill > 15) skill = 1
-      if (typeof burst !== "number" || burst < 1 || burst > 15) burst = 1
+      auto = typeof auto !== "number" ? 1 : clamp(auto, 1, 10)
+      skill = typeof skill !== "number" ? 1 : clamp(skill, 1, 10)
+      burst = typeof burst !== "number" ? 1 : clamp(burst, 1, 10)
```
nguyentvan7, line 47: Should we add logic based on ascension as well? No tracking issue at the moment, but its a known bug

apps/frontend/src/app/Database/DataManagers/CharacterData.ts
```diff
@@ -34,17 +34,17 @@ export class CharacterDataManager extends DataManager<CharacterKey, "characters"
...
 
     const { level, ascension } = validateLevelAsc(rawLevel, rawAscension)
 
     if (typeof talent !== "object") talent = { auto: 1, skill: 1, burst: 1 }
     else {
       let { auto, skill, burst } = talent
-      if (typeof auto !== "number" || auto < 1 || auto > 15) auto = 1
-      if (typeof skill !== "number" || skill < 1 || skill > 15) skill = 1
-      if (typeof burst !== "number" || burst < 1 || burst > 15) burst = 1
+      auto = typeof auto !== "number" ? 1 : clamp(auto, 1, 10)
```
nguyentvan7, line 45: Should we clamp to 12 for skill+burst and 14 for auto? Or we can just add "Bonus Normal Talent Level" + etc to the bonus stats page in a different PR

nguyentvan7 (changes requested):

apps/frontend/src/app/Database/DataManagers/CharacterData.ts
frzyc: Its not selectable via the dropdown, which can cause problem with UI. the recommended way to add "bonuses" are from bonus stats like childe

apps/frontend/src/app/Database/DataManagers/CharacterData.ts
frzyc: i thought about this, the issue is that if level/asc/talent is out of whack, which is the true value that everything is validated against? i'd argue the truth is level, which validates asc, which validates talent.

apps/frontend/src/app/Database/DataManagers/CharacterData.ts
nguyentvan7: #926

apps/frontend/src/app/Database/DataManagers/CharacterData.ts
nguyentvan7: Yea its not often for those things to be out of sync so asc or level is fine to validate first, and validate talent as a final step. Pretty rare that would affect anyone so no big deal in having it be slightly arbitrary imo

apps/frontend/src/app/Database/DataManagers/CharacterData.ts
frzyc: @nguyentvan7 is the expectation that I incorporate #926 here? I think changes here so far should be sufficient

PR #931, opened 2023-02-02
fix crash from artifacts with zero lines

frzyc:
Funky scanner issue, this is intended not to whitescreen the site if this happens. 
Will need to have discussion on whether a "zero line" artifact should be an "invalid" artifact, due to only 1*/2* artifact can have 0 lines, and currently low rarity artifacts are not supported in GO.


nguyentvan7 (approved):
image
it does make the filters technically not line up if you are to have 0-line artifacts. but this basically should never happen so Im fine with leaving it as-is

frzyc:
Yeah I wanna address the crash 1st, because user is unable to delete the funky artifacts

PR #941, opened 2023-02-04
Tweaks a few strings, tweaks some conditionals structures.

### LINKED ISSUE #937: Instructors set is using wrong string
Should just use the "trigger an elemental reaction" string in sheet.json

PolomoPT:
So uh the first commit is just the string tweaks. I do want to create a few and delete a few, though, maybe Van will help me with that if he bears with me on Discord.
Really just wording changes, not much. I'll probably describe any further changes on the commits themselves or maybe edit this.

Resolve #937


PolomoPT:
I'm gonna need maybe Van to change the string used by Faruzan in one of her burst conds to "baleCondName", if it's not too much trouble. Made a separate commit just in case.

PolomoPT:
So this one now would imply on making two conditionals instead of the current one. The 1st one would use "afterUse.burst" from sheet.json (representing the con's 1st effect), and when activated, would make another conditional appear, which would use the currently-used string, c4ReactCond, representing the con's 2nd effect. Explanations on why are in Discord. Again, separate commit just in case

PolomoPT:
For this commit to work, current uses of Sayu's "p1Swirl" need to be replaced with sheet.json's "energyRegen" and the "c4Ener" conditional needs to be removed entirely and display its effects plainly.

nguyentvan7:
I'm going to opt to keep Tighnari C4 the same as it is. Less churn for users and it accomplishes the same effect. I will take the string change but not implement the conditional change.

libs/localization/assets/locales/en/sheet.json
```diff
@@ -149,5 +151,16 @@
     "skill": "Skill DMG Mult."
   },
   "energy": "Energy",
-  "triggerQuota": "Trigger Quota"
+  "triggerQuota": "Trigger Quota",
+  "enemyAffected": {
```
frzyc, line 155: Not action item, but perhaps situations like this can be simplified using [nesting](https://www.i18next.com/translation-function/nesting)?

nguyentvan7 (approved):

frzyc (approved):

libs/localization/assets/locales/en/sheet.json
nguyentvan7: Probably worth doing here, these strings feel like a perfect fit. I'll see how hard it is and revise the pr

nguyentvan7: I added it. We could (or should?) migrate/consolidate all translations like this moving forward with interpolation maybe, instead of having a translator translate a string multiple times (like for hitOp). Although that means strings might become more awkward to accommodate such things.

One English example I can think of is if we had a generic string "After triggering a {element} reaction". We can easily interpolate "hydro", "cryo", etc in there and be fine. But to interpolate "electro" in there, we would need to modify the base string to be "After triggering a**(n)** {element} reaction", which is a bit clunky. I don't know enough about other languages to know if this is something that could occur in them as well, probably with masculine/feminine nouns+verbs etc.

But probably still worth consolidating to save translation effort? In the future though

frzyc: I think you can add in `a` vs `an` handling using [formatting](https://www.i18next.com/translation-function/formatting)

PR #948, opened 2023-02-05
Update repository readme and gitignore

nguyentvan7:


lantua (commented):
We should adopt this: https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored

frzyc (approved):

PR #953, opened 2023-02-06
frzyc jumped the gun now we need to have more stuff here

frzyc:


.gitattributes
```diff
@@ -0,0 +1,107 @@
...
+#
+
+# source code
+*.php text
+*.css text
+*.sass text
+*.scss text
+*.less text
+*.styl text
+*.js text eol=lf
```
nguyentvan7, line 22: `.ts` and `.tsx` found dead in a ditch

frzyc: LOL i just assumed they had .ts files

frzyc: Technically .ts and .tsx should be covered by `* text=auto`, I just switched between tooflesswolf's PR and this one and no funky LF files got generated

nguyentvan7 (approved):

.gitattributes
nguyentvan7: Seems like the rest of the text endings are unneeded, but I'm too lazy to figure it out

nguyentvan7:
also, the PR title should be "frzyc jumped the gun". You approved, a merge was inevitable :D

PR #967, opened 2023-02-11
Convert DropdownButton to use Popper-based Menu

### LINKED ISSUE #936: Changing character talent drop downs cause page to freeze
Only on the talent page, not on team-mate talent drop downs.

Seems to be isolated to my database as far as I can tell - but can reproduce it in Chrome and Edge using my database.

Steps:
Go to Raiden
Change her resolve stacks to something else

Also failing on Yoimiya changing her A1 passive dropdown

Interestingly doesnt happen for Yelan's A4

[Stain_2023-02-03_09-25-32.txt](https://github.com/frzyc/genshin-optimizer/files/10577510/Stain_2023-02-03_09-25-32.txt)

nguyentvan7:
* Resolve #936 

Convert DropdownButton to use a custom-created menu-like component that uses `Popper` instead of `Popover`. Difference is that `Popper` does not create an uninteractable overlay across the entire page, and also allows scrolling, which makes the site feel more natural.  
Worth noting that the race condition issue actually isn't fixed. If you look at the DOM, you can still see the Popper will get stuck sometimes after selecting something from the dropdown. However, you are still able to interact with the site, and the stuck dropdown is no longer interactable. So for the end user, the bug is fixed. The next time you interact with a menu, the previous Popper will clean itself up.

Component lifted from https://mui.com/material-ui/react-menu/#menulist-composition

### DIFF
--- a/apps/frontend/src/app/Components/DropdownMenu/DropdownButton.tsx
+++ b/apps/frontend/src/app/Components/DropdownMenu/DropdownButton.tsx
@@ -1,5 +1,5 @@
 import { KeyboardArrowDown } from "@mui/icons-material";
-import { Button, ButtonProps, Menu, Skeleton } from "@mui/material";
+import { Button, ButtonProps, ClickAwayListener, Grow, MenuList, Paper, Popper, Skeleton } from "@mui/material";
 import { Suspense, useCallback, useState } from "react";
 
 export type DropdownButtonProps = Omit<ButtonProps, "title"> & {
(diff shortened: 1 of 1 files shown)


frzyc (approved):
Yeah, I was actually trying to fix this by recoding MUI's implementation of `Select` but with a Button, but I think this is a good fix for now(since this fixes the undesired UI behaviour). Good summary on the `Poppver` vs `Popover` difference.

PR #975, opened 2023-02-12
Optimize runtime of tally counters

nguyentvan7:
Actually not sure if this will be noticeable. But the changes make sense to speed it up.


nguyentvan7:
Should probably explain my changes :)

* Minimize number of iterations over entire artifact database by calculating the various tallies in a single pass
* Pass tallies to different components in optimize tab rather than recalculate them, further reducing passes over the artifact db
* Convert filtered artifact ids to a map, so we don't have to check the filtered list for every single artifact ID when counting (O(n*m) monkaS)

libs/consts/src/artifact.ts
```diff
@@ -57,3 +57,5 @@ export const artMaxLevel: Record<RarityKey, number> = {
   4: 16,
   5: 20,
 } as const
+
+export const allArtifactRarityKeys = [5, 4, 3] as const
```
frzyc, line 61: This is a GO-only field because only GO limits artifact rarity. I would prefer if this stays in the front-end only side of the code

frzyc:
other than one nit, LGTM

PR #985, opened 2023-02-14
add sliders to artifact editor

### LINKED ISSUE #984: Adding an artifact manually pre-fills the lowest value for each substat as they're added
image

Meaning to change to a different initial roll value you have to now edit the number rather than clicking on the first value you want.

Could use this oppertunity to enhance the Artifact Editor by changing the buttons to a slider with the devisions set to valid substat rolls.

This would be a nice QOL especially for mobile users who have to manually edit the text box if accidently pressed the wrong value box.

frzyc:
image
Resolves #984.

### DIFF
--- a/apps/frontend/src/app/PageArtifact/ArtifactEditor/Components/SubstatInput.tsx
+++ b/apps/frontend/src/app/PageArtifact/ArtifactEditor/Components/SubstatInput.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, ButtonGroup, Grid, ListItemIcon, ListItemText, MenuItem, Slider, Typography } from '@mui/material';
+import { Box, Button, ButtonGroup, Grid, ListItemIcon, ListItemText, MenuItem, Slider, SliderProps, Typography } from '@mui/material';
 import { useEffect, useMemo, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import CardLight from '../../../Components/Card/CardLight';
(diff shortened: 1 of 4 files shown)


apps/frontend/src/app/PageArtifact/ArtifactEditor/Components/SubstatInput.tsx
```diff
@@ -37,6 +38,8 @@ export default function SubstatInput({ index, artifact, setSubstat }: { index: n
   if (!rollNum && key && value) error = error || t`editor.substat.error.noCalc`
   if (allowedRolls < 0) error = error || t("editor.substat.error.noOverRoll", { value: allowedRolls + rollNum })
 
+  const marks = useMemo(() => key ? artifactSubRolls(rarity, key).map(v => ({ value: v, })) : [{ value: 0 }], [key, rarity])
```
nguyentvan7, line 41: Should we have a 0-state as well? Otherwise you can't select an initial roll with the 4-button style input, as the 4 buttons always start with the second roll
image

frzyc: Yeah i was wondering about this. The issue is that the validation code is trying to match the min value, but the editor should allow for "non valid" values while user is editing. either i let validation allow a 0, or add a special cached value in the UI that allows user more freedom of editing(will get desynced from the actual underlying data representation)

nguyentvan7, line 41: You can't drag the slider back to 0 once you've moved it. Feels kind of unintuitive if you touch the slider but want to move it back so you can use the buttons to change the first roll

nguyentvan7 (commented):
If you really dont want the slider to go back to 0, then lgtm

apps/frontend/src/app/PageArtifact/ArtifactEditor/Components/SubstatInput.tsx
frzyc: yeah I want to discourage this behavour, since if they leave it at 0 it will actually get validated for min value when they save. they can always enter 0 in the input to go to back to that state

nguyentvan7 (approved):

PR #992, opened 2023-02-16
Add crash/recovery screen

nguyentvan7:
image
To test, you can toss this line in at line 96 in `ArtifactFilterDisplay.tsx` and then go to the Artifacts page

`        ctMap.locationTotal["crash"].total++`


apps/frontend/src/app/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+interface Props extends WithTranslation {
+  children?: ReactNode
+}
+
+interface State {
+  error?: Error
+}
+
+class ErrorBoundary extends Component<Props, State> {
```
frzyc, line 22: ew is this a class based component? what is this? 2015?

nguyentvan7: Ikr needs to be class based because I need to implement getDerivedStateFromError for error boundary

frzyc (approved):
LGTM

PR #944, opened 2023-02-04
Artifact & Character exclusion build settings are now per character

frzyc:
* Resolve #901 
* Fix #991 
image
image
image

PR #993, opened 2023-02-16
Standardize UseExcluded and UseEquipped tally display

### LINKED ISSUE #990: Use excluded button cannot be enabled even when artifacts are excluded
Need to increment counter before checking location status, I assume

nguyentvan7:
* Resolve #990


frzyc:
Probably this in #944, so I'm not sure if there is any point other than introducing a conflict i'll have to resolve later...

nguyentvan7:
How soon are you merging that? I could also bandaid fix the issue with the button being disabled, as it's currently completely broken and fix asap

nguyentvan7:
Closing to have mvp fix since you're going to rework exclusion and equipped logic to be per-char anyways so these counters will need to be redone as well

PR #1004, opened 2023-02-19
LP-based linearUpperBound calculator for BNB process.

tooflesswulf:
I left the existing linearUpperBound function in place b/c both are valid upper bounds. We can technically do $$UB = \min [ UB_{AMGM}, UB_{LP} ]$$ for a tighter bound, but I haven't done any testing for whether it's worth the computational cost.

Correctness is not fully tested, though it's matched live on the 5-ish characters I tested.


apps/frontend/src/app/Solver/GOSolver/solveLP_simplex.ts
```diff
@@ -0,0 +1,119 @@
+/**
```
lantua, line 1: Rename the file to `solveLP.ts`. We don't need to denote the algorithm in the file name as it could change without changing the APIs.

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
```diff
@@ -0,0 +1,80 @@
...
+    const prod = coords.reduce((prod, v) => prod * v, 1)
+    return [
+      [...coords.map(v => -v), 1, 0, -prod],
+      [...coords, -1, -1, prod],
+    ]
+  })
+
+  const objective = [...bounds.map(_ => 0), 0, 1]
+  try {
+    // TODO: verify solution
```
lantua, line 67: Does this still apply? If so, what do we need to verify?

apps/frontend/src/app/Solver/GOSolver/solveLP_simplex.ts
lantua, line 1: Rename the file to solveLP.ts. We don't need to denote the algorithm in the file name as it could change without changing the APIs.

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Monomial is assumed to be
+ *    m(x) = x1 * x2 * ... * xn
+ * on bounded domain
+ *    min_1 <= x1 <= max_1
+ *    min_2 <= x2 <= max_2
+ *    ...
+ *    min_n <= xn <= max_n
+ *
+ * @param bounds List of min/max bounds for each xi
+ * @returns A linear function L(x) = w•x + $c
```
lantua, line 43: Let's stick to ASCII (`*`) even if it is slightly less accurate.

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
```diff
@@ -0,0 +1,80 @@
...
+
+function weightedSum(...entries: readonly (readonly [number, Linear])[]): Linear {
+  const result = { $c: 0 }
+  for (const [weight, entry] of entries)
+    for (const [k, v] of Object.entries(entry))
+      result[k] = (result[k] ?? 0) + weight * v
+  return result
+}
+
+export function linearUB(nodes: OptNode[], arts: ArtifactsBySlot) {
```
lantua, line 16: This is exposed API, please add return type. If it is complex, make a new `type` declaration.

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
```diff
@@ -0,0 +1,80 @@
...
+  const nVar = bounds.length
+
+  // Re-scale bounds to [0, 1] for numerical stability.
+  const boundScale = bounds.map(({ max }) => max)
+  const scaleProd = boundScale.reduce((prod, v) => prod * v, 1)
+  bounds = bounds.map(({ min, max }) => ({ min: min / max, max: 1 }))
+
+  // Setting up the linear program in terms of constraints.
+  //   cartesian(bounds) loops 2^nVar times
+  const cons = cartesian(...bounds.map(({ min, max }) => [min, max])).flatMap((coords) => {
```
lantua, line 57: `.map.flatMap` => `flatMap`

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
```diff
@@ -0,0 +1,196 @@
...
+function slopePoint(slope: number, x0: number, y0: number, poly: Polynomial): Polynomial {
+  return addP(y0 - slope * x0, mulP(slope, poly))
+}
+function interpolate(x0: number, y0: number, x1: number, y1: number, poly: Polynomial, upper: boolean): Polynomial {
+  if (Math.abs(x0 - x1) < 1e-10)
+    return constP(upper ? Math.max(y0, y1) : Math.min(y0, y1))
+  return slopePoint((y1 - y0) / (x1 - x0), x0, y0, poly)
+}
+
+export function polyUB(nodes: OptNode[], arts: ArtifactsBySlot) {
```
lantua, line 54: Exposed API, add return type.

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
```diff
@@ -0,0 +1,196 @@
...
+  const minMaxes = new Map<OptNode, MinMax>()
+  forEachNodes(nodes, f => {
+    const { operation } = f
+    if (operation === "mul") minMaxes.set(f, { min: NaN, max: NaN })
+    switch (operation) {
+      case "mul": case "min": case "max": case "threshold": case "res": case "sum_frac":
+        f.operands.forEach(op => minMaxes.set(op, { min: NaN, max: NaN })); break
```
lantua, line 61: Unnecessary `break`

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
```diff
@@ -0,0 +1,196 @@
...
+    return constP(upper ? Math.max(y0, y1) : Math.min(y0, y1))
+  return slopePoint((y1 - y0) / (x1 - x0), x0, y0, poly)
+}
+
+export function polyUB(nodes: OptNode[], arts: ArtifactsBySlot) {
+  const minMaxes = new Map<OptNode, MinMax>()
+  forEachNodes(nodes, f => {
+    const { operation } = f
+    if (operation === "mul") minMaxes.set(f, { min: NaN, max: NaN })
+    switch (operation) {
```
lantua, line 59: Is this meant to be exhaustive? If so, add `default: assertUnreachable(operation)`

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
```diff
@@ -0,0 +1,196 @@
+import { customMapFormula, forEachNodes } from "../../Formula/internal";
+import { OptNode, allOperations } from "../../Formula/optimization";
+import { ConstantNode } from "../../Formula/type";
+import { prod, threshold } from "../../Formula/utils";
+import { assertUnreachable, cartesian } from "../../Util/Util";
+import { ArtifactsBySlot, MinMax, computeFullArtRange, computeNodeRange } from "../common";
+import { Linear } from "./BNBSplitWorker";
+
+export type Polynomial = PolyProd | PolySum | LinTerm
```
lantua, line 9: Add a brief description for each of these three

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
```diff
@@ -0,0 +1,196 @@
...
+type LinTerm = { type: 'lin', lin: Linear }
+type PolyProd = { type: 'prod', terms: Polynomial[], $k: number }
```
lantua, line 11: Can we switch to `$c` so that they are consistent, or is there a reason for the difference?

apps/frontend/src/app/Solver/GOSolver/solveLP_simplex.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export function solveLP(c: number[], Ab: number[][]) {
+  const rows = Ab.length + 1
+  const cols = Ab[0].length
+
+  const tableau = Array(rows).fill(0).map(_ => Array(cols).fill(0))
+  Ab.forEach((Ai, i) => Ai.forEach((Aij, j) => tableau[i][j] = Aij))
+  c.forEach((cj, j) => tableau[rows - 1][j] = cj)
+
+  const ijTrack: { i: number, j: number }[] = []
```
lantua, line 27: Is there a more descriptive name for this thing?

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
lantua, line 57: `.map.flatMap` => `.flatMap`

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
lantua, line 67: Does this still apply? If so, what do we need to verify?

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
lantua, line 43: Let's stick to ASCII (`*`) even if it is slightly less accurate.

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
lantua, line 54: Exposed API, please add return type

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
lantua, line 9: Add a brief description for each of these three

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
lantua, line 11: Can we switch to `$c` so that they are consistent with other types, or is there a reason for the difference?

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
```diff
@@ -0,0 +1,196 @@
...
+      default: assertUnreachable(operation)
+    }
+  })
+
+  return poly.map(p => expandPoly(p))
+}
+
+export type SumOfMonomials = Monomial[]
+type Monomial = {
+  $k: number
```
lantua, line 142: Can we switch to `$c` so that they are consistent with other types, or is there a reason for the difference?

apps/frontend/src/app/Solver/GOSolver/solveLP_simplex.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * Solve a Linear Program defined by:
```
lantua, line 2: There are some places where `i` and `j` seems to follow some convension (`i` for row, `j` for column?). If so, add a note for that at the beginning of the file. Maybe switch to `r` and `c` to make it more "intuitive"?

apps/frontend/src/app/Solver/GOSolver/solveLP_simplex.ts
lantua, line 1: Rename the file to `solveLP.ts`. We shouldn't denote the algorithm in the file name as it could change without changing the APIs

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
lantua, line 16: This is exposed API, please add return type. If it is complex, make a new type declaration.

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
lantua, line 61: Unnecessary `break`

lantua (changes requested):
Aside from the comments, could you
- check if you need the old `linearUpperBound`, and if not, remove them,
- add some tests for `polyUB` and `solveLP`?

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
tooflesswulf: do you accept `@` as inner product symbol? If not imma do `w . x` because I typically write `w * x` to mean element-wise product

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
tooflesswulf: I intended `$c` to be additive constant (initializes to 0) and `$k` to be multiplicative constant (initializes to 1)

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
tooflesswulf: No, it needs to be this way. Parentheses are like:

```js
cartesian(...)
  .flatMap((coords) => { /* stuff */ })
```

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
tooflesswulf: Just need to verify the solution satisfies the constraints. I've implemented it in `solveLP` so it checks the solution is valid before returning it.

apps/frontend/src/app/Solver/GOSolver/solveLP_simplex.ts
tooflesswulf: renamed to pivotRecord

lantua: Is this the term used in the referenced link? if not, please provide a brief description.

apps/frontend/src/app/Solver/GOSolver/linearUB.ts
lantua: Either should be fine, or we can just use `dot(w, x)` if we don't have a commonly used notation.

apps/frontend/src/app/Solver/GOSolver/polyUB.ts
lantua: We should note that difference somewhere. Maybe at the top of a file where these are used the most?

lantua (changes requested):
Please add tests and merge changes from `master`.

apps/frontend/src/app/Formula/internal.ts
```diff
@@ -88,7 +88,7 @@ export function customMapFormula<Context, Output, Input extends Base<Input>>(for
     const [visiting, mapping] = current
 
     const old = mapping.get(formula)
-    if (old) return old
+    if (old) return { ...old }
```
lantua, line 91: Don't create new objects; it'll decache the results.

tooflesswulf: Oh lmao I did this cuz I found the cached objects were being modified, breaking the bounding alg. How can I change customMapFormula to return a read-only object?

lantua: We unfortunately does not have guardrails to prevent mutation, but immutability is the foundational assumption of waverider. `customMapFormula` itself does not mutate a `Node`, nor do any other existing functions. There are only a handful of exceptions, but they are limited to when it is guaranteed that the mutation is not observable.

Make sure there is no mutation in your code. If the mutation comes from existing codebase, please file a bug report.

lantua (approved):
We can't realistically test that `linearUB` will return on every well-formed arguments, especially around degeneracy, but this should be sufficient testing. Tests may also need some churning after WR2 due to how it it constructed.

With that said, this should be sufficient for merging. If we found cases where `linearUB` throws in normal usage, we may need to tweak or change the LP algorithm.

PR #921, opened 2023-01-29
update dependency, fix webworker

frzyc:
Manually fix webpack issues:
- Webworkers not working on local dev env
- assets not loading with latest nx webpack
- hrefbase not working on deployment to github pages

Still in draft:
- tree shaking is not working

PR #1010, opened 2023-02-21
update nx, fix treeshaking, fix local webworker

frzyc:
A updated version of #921
With some changes to the Webpack config to 

1. Allow local web workers to work without that `import.meta` err
2. Add tree shaking
3. Add minification

Also, no more cmd windows popping up on windows.


apps/frontend/project.json
```diff
@@ -59,7 +59,7 @@
           "namedChunks": false,
           "extractLicenses": true,
           "vendorChunk": false,
-          "baseHref": "/genshin-optimizer/"
+          "baseHref": "/genshin-optimizer-beta/"
```
nguyentvan7, line 62: Probably not intended

apps/frontend/project.json
```diff
@@ -83,7 +83,7 @@
       "dependsOn": ["^build", "^pipeline"]
     },
     "deploy":{
-      "command": "gh-pages -d dist/apps/frontend",
+      "command": "gh-pages -d dist/apps/frontend -r https://github.com/frzyc/genshin-optimizer-beta",
```
nguyentvan7, line 86: Ditto

nguyentvan7 (changes requested):

apps/frontend/project.json
frzyc: oh yeah, i was testing deployment against beta site.

apps/frontend/src/polyfills.ts
```diff
@@ -1,7 +0,0 @@
-/**
```
nguyentvan7, line 1: Does this mean we aren't using polyfills? or it's just used in a different way?

apps/frontend/webpack.config.js
```diff
@@ -0,0 +1,27 @@
+const { composePlugins, withNx } = require('@nrwl/webpack');
+const { withReact } = require('@nrwl/react');
+const TerserJSPlugin = require('terser-webpack-plugin');
+
+// Nx plugins for webpack.
+module.exports = composePlugins(withNx(), withReact(), (config) => {
+  // Update the webpack config as needed here.
+  // e.g. `config.plugins.push(new MyPlugin())`
+
+  // https://github.com/nrwl/nx/issues/10017#issuecomment-1405429851
```
nguyentvan7, line 10: Can you leave a comment as to why we added this
```suggestion
  // https://github.com/nrwl/nx/issues/10017#issuecomment-1405429851
  // Fix to web workers not working
```

apps/frontend/webpack.config.js
```diff
@@ -0,0 +1,27 @@
...
+  config.output.scriptType = "text/javascript"
+
+
+  // https://github.com/nrwl/nx/issues/14483#issuecomment-1424029390
+
+  // Fix 1: resolves tree-shaking issue
```
nguyentvan7, line 16:
```suggestion
  // Fix 1: resolves tree-shaking issue caused by using custom webpack config
```

apps/frontend/src/polyfills.ts
frzyc: When i generated a new app, and look at their config, they dont include the polyfill anymore, so either its deprecated, or... in theory webpack does the poly fill already, and this is just a override that we are not using

nguyentvan7 (approved):

PR #1019, opened 2023-02-23
fix art exclusion not excluding

frzyc:
Looks like logic was missing in #944


nguyentvan7 (approved):
Kinda tipsy, let's just assume I tested it

PR #1020, opened 2023-02-23
Add waverider 2

lantua:
This adds a new implementation of waverider engine as a separate module. It is not attached to the application, but the API has been stable for quite some time, so we should be able to safely add this to `master`.


nx.json
```diff
@@ -5,7 +5,7 @@
     "default": {
       "runner": "nx/tasks-runners/default",
       "options": {
-        "cacheableOperations": ["build", "lint", "test", "e2e"]
+        "cacheableOperations": ["build", "lint", "test", "e2e", "genfiles"]
```
frzyc, line 8: is `genfiles` cacheable? usually generated files are not even tracked, or tracked outside of the module they are generated from/to, which would prevent the caching of their opertaions?

frzyc: https://nx.dev/concepts/how-caching-works

frzyc (approved):

libs/waverider/src/node/calc.ts
```diff
@@ -0,0 +1,120 @@
...
+              return computed[accu]!
+          }
+        }
+        default: assertUnreachable(op)
+      }
+    }
+  }
+
+  computeMeta(_op: Exclude<AnyOP, 'read'>, _tag: Tag | undefined, _value: any, _x: (CalcResult<any, M> | undefined)[], _br: CalcResult<any, M>[], _ex: any): M {
+    return undefined as any
```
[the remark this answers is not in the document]
lantua: It defaults to `undefined`, and subclasses can implement it with custom logic.

nguyentvan7 (approved):

PR #1034, opened 2023-02-25
Add favorite indicator to CharacterCardPico

nguyentvan7:
image

### DIFF
--- a/apps/frontend/src/app/Components/Character/CharacterCardPico.tsx
+++ b/apps/frontend/src/app/Components/Character/CharacterCardPico.tsx
@@ -1,5 +1,7 @@
 import { CharacterKey } from '@genshin-optimizer/consts';
 import { characterAsset } from '@genshin-optimizer/g-assets';
+import FavoriteIcon from '@mui/icons-material/Favorite';
+import FavoriteBorderIcon from '@mui/icons-material/FavoriteBorder';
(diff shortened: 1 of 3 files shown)


frzyc:
Why is there so much code churn just to add an icon?

nguyentvan7:
UseCharMeta requires a characterkey !== "", Which char card pico does not guarantee. Think the refactor is better anyways right, since there's less unnecessary hook usage for the blank char card

nguyentvan7:
UseCharMeta requires a characterkey !== "", Which char card pico does not guarantee. Think the refactor is better anyways right, since there's less unnecessary hook usage for the blank char card

apps/frontend/src/app/Components/Character/CharacterCardPico.tsx
```diff
@@ -16,13 +20,29 @@ import CharacterCard from './CharacterCard';
 
 export default function CharacterCardPico({ characterKey = "", index = -1, onClick, simpleTooltip = false, }:
```
frzyc, line 21: I think in this case, this might be over-nested. I think `CharacterCardPico` should have a strict `character:CharacterKey` prop, and you alternatively export a separate `EmptyCharCardPico` with `index`. This prevents prop drilling, and enforces typing.

frzyc (changes requested):

frzyc (approved):

PR #1035, opened 2023-02-25
Allow click-dragging to select in the priority list

nguyentvan7:
chrome_wldx7G8zNh

This is only on desktop. Mobile will only keep tap functionality to avoid messing with scrolling


frzyc:
Even in your demo this messed with tooltips. The issue is that when you hold onto any particular icon, it will trigger the delayed tooltip hover. you might need to find a way to disable the tooltip when it is registering a hold.
Does this toggle or does each drag only allow/disallow?

nguyentvan7:
the drag will toggle everything you dragged over. The second drag of the example shows that a bit

nguyentvan7:
I added disabling the tooltip. You can also try commit 1eb12ca, as adjusting the timing makes it a lot better. If the user is taking more than 0.5s to click-drag out of the current character, feels like theyre a bit slow, but I will leave it up to you

apps/frontend/src/app/Components/Character/CharacterCardPico.tsx
```diff
@@ -30,7 +30,7 @@ export default function CharacterCardPico({ characterKey, onClick, simpleTooltip
   )
 
   const simpleTooltipWrapperFunc = useCallback((children: ReactNode) =>
-    <BootstrapTooltip placement="top" title={
```
frzyc, line 33: this variant should have no deplay, since its just a simple name tooltip that goes in the main tooltip

frzyc (changes requested):
image

nguyentvan7:
I wonder if being able to toggle everything is too complex. Like if the user starts a drag on an excluded character, should that drag only select excluded characters until they finish?

nguyentvan7:
Refer to https://github.com/frzyc/genshin-optimizer/pull/1068

PR #1052, opened 2023-03-02
3.5 Polomo Pull Part ' + Tiny Textual Tweaks

PolomoPT:
'Tis the thing to do.

### DIFF
--- a/libs/localization/assets/locales/en/char_Dehya.json
+++ b/libs/localization/assets/locales/en/char_Dehya.json
@@ -3,6 +3,5 @@
   "field_dmg_": "Fiery Sanctum DMG Bonus",
   "initialHeal": "Initial Healing",
   "dotHeal": "Additional Healing",
-  "critHitDuringBurst": "Number of CRIT hits during a single use of <strong>The Lioness' Bite</strong>",
-  "onHitHeal": "On-hit Healing"
+  "critHitDuringBurst": "After Flame-Mane's Fist deals a CRIT hit"
 }
--- a/libs/localization/assets/locales/en/char_TravelerGeo.json
+++ b/libs/localization/assets/locales/en/char_TravelerGeo.json
@@ -4,5 +4,8 @@
   },
   "c2": {
     "key": "Rockcore Meltdown DMG"
+  },
+  "c4": {
+    "energyRestore": "Energy restored per shockwave hit"
   }
 }
--- a/libs/localization/assets/locales/en/char_TravelerGeo.json
+++ b/libs/localization/assets/locales/en/char_TravelerGeo.json
@@ -4,8 +4,5 @@
   },
   "c2": {
     "key": "Rockcore Meltdown DMG"
-  },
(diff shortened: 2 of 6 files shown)


libs/localization/assets/locales/en/char_Dehya.json
```diff
@@ -3,6 +3,5 @@
   "field_dmg_": "Fiery Sanctum DMG Bonus",
   "initialHeal": "Initial Healing",
   "dotHeal": "Additional Healing",
-  "critHitDuringBurst": "Number of CRIT hits during a single use of <strong>The Lioness' Bite</strong>",
-  "onHitHeal": "On-hit Healing"
```
frzyc, line 7: Why are we removing this? we dont have an occurrence of this string in the system.

PolomoPT:
Changing to regular Healing, because it works the same

Em qui, 2 de mar de 2023 17:06, frzyc ***@***.***> escreveu:

libs/localization/assets/locales/en/char_TravelerGeo.json
```diff
@@ -4,8 +4,5 @@
   },
   "c2": {
     "key": "Rockcore Meltdown DMG"
-  },
-  "c4": {
-    "energyRestore": "Energy restored per shockwave hit"
```
frzyc, line 9: Why are we removing this to use a more vague text?

PolomoPT:
Because it's standardizing and what is done with other
constellations/talents that provide simillar effects. e.g. venti's burst
doesn't say "energy restored upon burst ending"

Em qui, 2 de mar de 2023 17:09, frzyc ***@***.***> escreveu:

libs/localization/assets/locales/en/char_Dehya.json
PolomoPT: More vague doesn't mean worse. It's standardizing.

libs/localization/assets/locales/en/char_TravelerGeo.json
PolomoPT: Other characters also use simply "Energy Regeneration". Venti's Burst, for example, doesn't say "Energy restored upon Burst ending", it just uses the standardized string. That's the point of having Sheet.

frzyc: I am all for standardization, but not at the cost of context. I think its good to prompt user on how/why a condition occurs, (and use a manually translated string) than to create a generalized standardized string that gives less information. Especially since in the case of geo traveler, this string has already been translated to several languages already, I see this as a net loss to use a generalized string.

PolomoPT: that's valid, ig. same for mona, y'reckon?

frzyc: The Mona string is fine.

frzyc (approved):

PR #1058, opened 2023-03-04
Revamp behavior of exclude and allow list

### LINKED ISSUE #1026: Change "Exclude Artifacts" to "Excluded Artifacts"
Using the word "Exclude" communicates that the main purpose of clicking this is to add artifacts to the exclude list.
Where it's main purpose is to view, delete or "manage" the artifacts that are already excluded (since while adding an artifact is possible via this list, it's awkward)

image

image

### LINKED ISSUE #1027: Change "Allow Equipped Artifacts" back to a checkbox and change function slightly
The current wording for this box communicates that clicking it "allows" the use of any equipped artifacts.
Where infact it just opens a popup where the default state is still _NOT_ to use equipped artifacts.

image

This introduces two issues;
1) misunderstanding of the feature
2) any new character added to the database after configuring this popup will default to the "disallowed" state

The second being the major concern and inconvience as users will have to go back to every single character to allow any newly added character.

My idea would be:
Change the button back to a check box.
Default state: off (not allowing the use of equipped artifacts)
Clicked state: on (allowing the use of _ALL_ equipped artifacts by default; unless configured via the popup)

Re-introduce the settings cog to open the newly implemented popup but it now defaults _ALL_ characters (including new characters) to the "allowed" state.

### LINKED ISSUE #1028: This giant blue button is ugly
image

I hate it. Bring back the cog

peeporiot

### LINKED ISSUE #1029: Move (or remove) "Allow Partial Builds" button
image

I'm sure you've left this here on purpose just to trigger my OCDs and I can't take it anymore.

Put it at the bottom instead of in the middle of all these blue buttons.

pepeagony

### LINKED ISSUE #1033: Now unable to quickly allow a character to use equipped and then return them to their previous setting
In the prio list feature you could quickly add a character or move an existing one to the top of the list for testing purposes.
Then when you were done you could either delete that single character or move them back to their orginal position via the text box input.

Now this is no longer an easy process. Say for example I have the following equipped status setup on Rosaria:
image

I can easily click "add all to allow list" to achieve the first part of the goal. However setting this state up again is now a much more difficult process since the rest of the list has been wiped and no longer being "held" at a lower prio state or as a visual reference to slot the character back into their previous prio queue.

Issue #1030 would help this to a degree. But there is no guarantee that the "default" state the user has configured will be the state they had, or wanted this particular character to be in.

Currently I would have to memorize this equipped list and manually recreate it after doing the testing I wanted to do.
Another option would be creating a second copy of this database to run tests in which I can then just delete after so the "main" database is left untouched. Eitherway neither options are ideal and feel like a downgrade on the functionality the prio list had previously.

Issue #401 may be a solution if you could create another "instance" of the character for testing a different team or conditional state in so you don't have to destroy your current "loadout" settings. But there still may be times you want to quickly see what the "best" build option your character has compared to their limited selection option is without completely wiping the equipped allow/disallow list.

nguyentvan7:
- [x] Resolve #1026 
- [x] Resolve #1027 
- [x] Resolve #1028 
- [x] Resolve #1029
- [x] Resolve #1033
- [x] Add filter settings (no need to retain them) to the allowlist
- [x] Turn custom allow list behavior to be default-allow for any new characters. Lists will start with all-allow, if they were not configured before this change.
image
image


frzyc:
Probably need to deploy to a beta site and get stain to use it. so if people have problem with it after we release, we can blame stain for bad QA

nguyentvan7:
Refer to #1068

PR #1063, opened 2023-03-05
Add patch notes to home screen

nguyentvan7:
This is just a silly amount of packages to render some markdown, so I'm fine if this gets rejected for that.

There is also the chance we get rate limited since there is no API key. Idk what the limits are or really how many people would even be sending requests, so I don't think it'll be too much of an issue

image


apps/frontend/src/app/PageHome/index.tsx
```diff
@@ -49,3 +55,29 @@ function IntroCard() {
     </CardContent>
   </CardDark>
 }
+
+function PatchNotesCard() {
+  const { t } = useTranslation("page_home")
+  const [{ isLoaded, text }, setState] = useState({ isLoaded: false, text: "" })
+  console.log(process.env.NX_URL_GITHUB_API_GO_RELEASES)
```
frzyc, line 62: extra log

frzyc (approved):

PR #1067, opened 2023-03-06
Add functional API to database's set functions.

frzyc:
Generally in current system, you have to effectively keep both the value and setter, if a new value depends on a previous value. Now you can just use the setter, and not maintain a value.

This should reduce some UI cases where a button only sets a value without reading it, but both value and setter is funneled in, causing extra updates when value changes.

### DIFF
--- a/apps/frontend/src/app/Database/DataEntry.ts
+++ b/apps/frontend/src/app/Database/DataEntry.ts
@@ -24,31 +24,45 @@ export class DataEntry<
     this.key = key
     this.init = init
     this.goKey = goKey
-    if (this.database.storage.keys.includes(key)) {
-      this.data = this.toCache(init(this.database))!
-      this.set(this.database.storage.get(key))
-    } else {
-      this.data = this.toCache(init(this.database))!
-      this.setCached(this.data)
-    }
+    this.set(
+      this.database.storage.keys.includes(key)
+        ? this.getStorage()
+        : init(this.database)
+    )
+    this.data = this.get() //initializer
   }
 
   listeners: Callback<CacheValue>[] = []
   get() {
     return this.data
   }
-  validate(obj: any): StorageValue | undefined {
+  validate(obj: unknown): StorageValue | undefined {
     return obj as StorageValue | undefined
   }
   toCache(storageObj: StorageValue): CacheValue | undefined {
-    return { ...storageObj } as any as CacheValue
+    return { ...storageObj } as unknown as CacheValue
   }
   deCache(cacheObj: CacheValue): StorageValue {
-    const { ...storageObj } = cacheObj as any
-    return storageObj as any as StorageValue
+    const { ...storageObj } = cacheObj
+    return storageObj as unknown as StorageValue
   }
-  set(value: Partial<StorageValue>): boolean {
-    const old = this.data
+  getStorage(): StorageValue {
+    return this.database.storage.get(this.key)
+  }
+  set(
+    value:
+      | Partial<StorageValue>
+      | ((v: StorageValue) => Partial<StorageValue> | void)
+  ): boolean {
+    const old = this.getStorage()
+    if (typeof value === 'function') {
+      if (!old) {
+        this.trigger('invalid', value)
+        return false
+      }
+      const temp = value(old)
+      value = temp ? temp : old
+    }
     const validated = this.validate({ ...old, ...value })
     if (!validated) {
       this.trigger('invalid', value)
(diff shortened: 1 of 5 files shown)


apps/frontend/src/app/Database/DataEntry.ts
```diff
@@ -24,31 +24,45 @@ export class DataEntry<
...
-    const old = this.data
+  set(
+    value:
+      | Partial<StorageValue>
+      | ((v: StorageValue) => Partial<StorageValue> | void)
+  ): boolean {
+    const old = this.getStorage()
+    if (typeof value === 'function') {
+      if (!old) {
+        this.trigger('invalid', value)
```
nguyentvan7, line 60: Why do we only trigger if its a function?

nguyentvan7: Is it ok to trigger on a function?

apps/frontend/src/app/Database/DataEntry.ts
```diff
@@ -24,31 +24,45 @@ export class DataEntry<
...
-  set(value: Partial<StorageValue>): boolean {
-    const old = this.data
+  set(
+    value:
+      | Partial<StorageValue>
+      | ((v: StorageValue) => Partial<StorageValue> | void)
+  ): boolean {
+    const old = this.getStorage()
+    if (typeof value === 'function') {
+      if (!old) {
```
nguyentvan7, line 59: Why do we ignore this path if storage doesn't exist?

apps/frontend/src/app/Database/DataEntry.ts
```diff
@@ -24,31 +24,45 @@ export class DataEntry<
...
+        this.trigger('invalid', value)
+        return false
+      }
+      const temp = value(old)
```
nguyentvan7, line 63: Can we just get rename `value` to `valueOrFunc` or similar.
Then if we know `valueOrFunc` is a function, can we store it as `const valueFunc = valueOrFunc`
The readability of this is terrible

apps/frontend/src/app/Database/DataEntry.ts
```diff
@@ -81,8 +97,8 @@ export class DataEntry<
     this.database.storage.set(this.key, this.deCache(this.data))
   }
 
-  trigger(reason: TriggerString, object?: any) {
-    this.listeners.forEach((cb) => cb(reason, object))
+  trigger(reason: TriggerString, object?: unknown) {
```
nguyentvan7, line 100: Why can't this be typed as `CacheValue`

apps/frontend/src/app/Database/DataEntry.ts
```diff
@@ -24,31 +24,45 @@ export class DataEntry<
...
     this.init = init
     this.goKey = goKey
-    if (this.database.storage.keys.includes(key)) {
-      this.data = this.toCache(init(this.database))!
-      this.set(this.database.storage.get(key))
-    } else {
-      this.data = this.toCache(init(this.database))!
-      this.setCached(this.data)
-    }
+    this.set(
```
nguyentvan7, line 27: I feel like this would read better as an if-statement, but that's personal choice. I just find this whole thing hard to read so that might be a me-problem

apps/frontend/src/app/Database/DataEntry.ts
frzyc: Functional API uses the "previous" value as the parameter for the callback, if there is no previous value, the callback shouldn't be called at all.

apps/frontend/src/app/Database/DataEntry.ts
frzyc: It should be fine, since we are not really reading the actual values of the event listener in this context, the `invalid` events are currently only used by the import pipeline, which doesn't use the functional API.

apps/frontend/src/app/Database/DataEntry.ts
frzyc: In invalid cases we are not passing in a "valid" `CacheValue`

nguyentvan7: Doesn't that make this an incorrect type assertion?

apps/frontend/src/app/Database/DataEntry.ts
nguyentvan7: seems weird but i don't know enough to say anything else

nguyentvan7 (commented):
Left a comment on the type assertion

apps/frontend/src/app/Database/DataEntry.ts
frzyc: yes. need to have an overhaul of the event trigger system to have "correct types", which isnt the scope of this PR.

Issue #1060, opened 2023-03-05
Setup Github actions


eeeqeee:
Automatically check:
- [x] ESlint
- [ ] TSC
  - [x] Libs
  - [ ] App
- [x] Tests
- [x] Prettier

Deploy:
- [ ] Auto builds for each pr (needs separate subdomain)
- [ ] Nightly build for master branch
- [x] 1 click deploy for new versions

nguyentvan7:
We would also need to set up prettier. Which also means embracing multi-line imports which i hate, but am willing to live with

PR #1069, opened 2023-03-07
Setup CI

eeeqeee:
See #1060
Prettier blows up the size of the import sections frequently so we might not want to add it


frzyc (approved):

frzyc (commented):
Why is `.gitattributes` removed?

eeeqeee:
Prettier can enforce no CRLF and the `.gitattributes` was making it diff against previous versions.

frzyc (approved):

PR #1070, opened 2023-03-07
Add useGlobalError hook, replace worker failure error

nguyentvan7:


apps/frontend/src/app/ReactHooks/useGlobalError.tsx
```diff
@@ -0,0 +1,18 @@
...
+When this callback is used somewhere within an error boundary,
+it will trigger it, even if you are in an async block
+
+```
+const throwError = useGlobalError()
+async function foo() { throwError(new Error()) }
+```
+*/
+export default function useGlobalError(): (e: Error) => void {
+  const [, setError] = useState()
```
frzyc, line 14: why is this `useState` necessary?

nguyentvan7: this forces the error to be thrown on the synchronous side of react so it can be caught by the error boundary

https://medium.com/trabe/catching-asynchronous-errors-in-react-using-error-boundaries-5e8a5fd7b971

https://github.com/facebook/react/issues/14981

frzyc (approved):

PR #1071, opened 2023-03-08
Hi Lantua

nguyentvan7:


lantua (changes requested):
`Counter` doesn't seem to behave correctly. Instead, it looks like a lousy implementation from some bad tutorial websites.

PR #1072, opened 2023-03-09
Connect error from background worker to ui side

lantua:
This makes sure that errors thrown from any worker triggers cancellation at `GOSolver`. It used to silently ignore error if thrown from the event handler.


nguyentvan7:
So do we need #1070 or nah

lantua:
we still need it.

PR #1075, opened 2023-03-13
Split solvers into 2 classes

tooflesswulf:
Refactored `GOSolver` -> `BNBSolver` and `EnumerateSolver` with no additional changes

Reason is to prepare for the algorithm to spend most of its time in SplitWorker rather than ComputeWorker, which may warrant changes to message structure.


tooflesswulf:
Not doing this

PR #1077, opened 2023-03-13
minor changes to CI/pipeline

frzyc:
- Make the generated asset index files from `g-assets-gen` more `prettier`
- Amend CI to use the `master` branch as the reference branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,4 +62,4 @@ jobs:
           yarn install --immutable --immutable-cache
       - name: Check Formatting
         run: |
-          yarn run nx format:check
+          yarn run nx format:check --base=master
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,4 +62,4 @@ jobs:
           yarn install --immutable --immutable-cache
       - name: Check Formatting
         run: |
-          yarn run nx format:check --base=master
+          yarn run nx format:check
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -62,4 +62,4 @@ jobs:
           yarn install --immutable --immutable-cache
       - name: Check Formatting
         run: |
-          yarn run nx format:check
+          yarn run nx format:check --base=master
```
eeeqeee, line 65: the action doesnt checkout `master`, so this is the same as before
CI minutes are free here, so we might as well keep checking all files with `yarn run nx format:check --all` to be cautious against extra CRLFs or other issues creeping

nx.json
```diff
@@ -56,5 +56,8 @@
         "unitTestRunner": "none"
       }
     }
+  },
+  "affected": {
+    "defaultBase": "origin/master"
```
eeeqeee, line 61: just `master` is better, since the default is `main` without mentioning the remote

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "frontend": "nx serve frontend",
     "deploy": "nx deploy frontend",
+    "format-write": "nx format:write --base=master",
```
eeeqeee, line 9: if we set the default to `master` this is unnecessary

eeeqeee (changes requested):

eeeqeee (approved):

PR #1081, opened 2023-03-14
Make sure `BackgroundWorker` yields during `split`

lantua:
The previous implementation used a resolved promises, which got coerced into sync. Yielding is necessary for performance reason as the solver may broadcast a threshold while the splitting process is ongoing. Incorporating these thresholds early can help with splitting efficiency/performance.

### DIFF
--- a/apps/frontend/src/app/Solver/GOSolver/BackgroundWorker.ts
+++ b/apps/frontend/src/app/Solver/GOSolver/BackgroundWorker.ts
@@ -25,7 +25,7 @@ async function handleEvent(e: MessageEvent<WorkerCommand>): Promise<void> {
         data.maxIterateSize
       )) {
         postMessage({ command: 'iterate', filter })
-        await Promise.resolve() // in case a `threshold` is sent over
+        await new Promise((r) => setTimeout(r)) // in case a `threshold` is sent over
       }
       break
     case 'iterate':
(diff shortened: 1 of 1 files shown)


apps/frontend/src/app/Solver/GOSolver/BackgroundWorker.ts
```diff
@@ -25,7 +25,7 @@ async function handleEvent(e: MessageEvent<WorkerCommand>): Promise<void> {
         data.maxIterateSize
       )) {
         postMessage({ command: 'iterate', filter })
-        await Promise.resolve() // in case a `threshold` is sent over
+        await new Promise((r) => setTimeout(r)) // in case a `threshold` is sent over
```
frzyc, line 28: what does this even do?

lantua: `Promised.resolved` returns a resolved promised, which may transparently turn into a sync computation during async-to-promise transformation. This change ensures that promise resolution happens on the next tick (and not immediately), forcing the compiler to treat that as an async computation.

frzyc (approved):

apps/frontend/src/app/Solver/GOSolver/BackgroundWorker.ts
lantua: After more experiments, the difference between tasks and microtasks is the culprit. Tasks are executed in the next event loop, while microtasks are executed before the next task.

https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide#tasks_vs._microtasks

frzyc (approved):

PR #1085, opened 2023-03-17
Add new team members to home page

frzyc:
- [x] Update new/existing profile/title/subtitle
- [x] Update layout to be 3 rows?
- [x] compress the images? they won't be larger than like 250x250

image


JoseZehi (approved):
`reens.png` seems to be kinda big compared to all the other files, but I don't think it's significant enough, right?

frzyc:
> `reens.png` seems to be kinda big compared to all the other files, but I don't think it's significant enough, right?

should be fine. its not significantly bigger. i think one of the images was like 1k x 1k and was like 1mb LOL

Jomata (approved):

PR #1093, opened 2023-03-21
New release action

eeeqeee:
Partial resolution of #1060

### DIFF
--- a/.github/workflows/new-release.yml
+++ b/.github/workflows/new-release.yml
@@ -1,0 +1,66 @@
+name: New Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version'
+        type: string
+        required: true
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+        cache: 'yarn'
+    - name: Get version # TODO(eeeqeee) add option to increment from current version
+      id: get-version
+      run: |
+        echo "version=${{ inputs.version }}" >> $GITHUB_OUTPUT
+    - name: Set git config
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+    - name: Change version
+      run: |
+        TEMP_FILE=$(mktemp)
+        jq --arg v "${{ steps.get-version.outputs.version }}" '.version = $v' package.json > $TEMP_FILE
+        cp $TEMP_FILE package.json
+        git add .
+        git commit -m "${{ steps.get-version.outputs.version }}"
+    - name: Yarn install
+      run: |
+        yarn install --immutable --immutable-cache
+    - name: Setup .env
+      run: |
+        echo $ENVFILE > apps/frontend/.env.local
+      env:
+        ENVFILE: ${{ secrets.ENVFILE }}
+    - name: Build genshin-optimizer
+      run: |
+        NX_URL_GITHUB_GO_CURRENT_VERSION="https://github.com/${{ github.repository }}/commit/$(git rev-parse HEAD)"
+        yarn run nx run frontend:build:production
+    - name: Deploy to Github Pages
+      run: |
+        git fetch origin gh-pages --depth=1
+        git worktree add ../gh-pages gh-pages --no-checkout
+        cp -rT dist/apps/frontend ../gh-pages
+        cd ../gh-pages
+        git add --all .
+        git commit -m "deploy: version ${{ steps.get-version.outputs.version }} $(git rev-parse HEAD)"
+        git push
+    - name: Push
+      run: |
+        git status
+        git push
+    - name: Make release
+      run: |
+        gh release create ${{ steps.get-version.outputs.version }} --generate-notes --target $(git rev-parse HEAD)
+      env:
+        GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/new-release.yml
+++ b/.github/workflows/new-release.yml
@@ -13,54 +13,54 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-    - name: Checkout
-      uses: actions/checkout@v3
-    - uses: actions/setup-node@v3
-      with:
-        node-version: 18
-        cache: 'yarn'
-    - name: Get version # TODO(eeeqeee) add option to increment from current version
-      id: get-version
-      run: |
-        echo "version=${{ inputs.version }}" >> $GITHUB_OUTPUT
-    - name: Set git config
-      run: |
-        git config user.name "github-actions[bot]"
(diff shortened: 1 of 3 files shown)


.github/workflows/new-release.yml
```diff
@@ -0,0 +1,66 @@
...
+      run: |
+        yarn install --immutable --immutable-cache
+    - name: Setup .env
+      run: |
+        echo $ENVFILE > apps/frontend/.env.local
+      env:
+        ENVFILE: ${{ secrets.ENVFILE }}
+    - name: Build genshin-optimizer
+      run: |
+        NX_URL_GITHUB_GO_CURRENT_VERSION="https://github.com/${{ github.repository }}/commit/$(git rev-parse HEAD)"
```
nguyentvan7, line 47: So does this somehow get put into `process.env.NX_URL_GITHUB_GO_CURRENT_VERSION` in our `tsx` files?

If so, doesn't that mean the footer will now link to latest commit, rather than latest patch notes?

Or is the footer change for future "deploy per-pr/nightly" action?

.github/workflows/new-release.yml
```diff
@@ -0,0 +1,66 @@
...
+        node-version: 18
+        cache: 'yarn'
+    - name: Get version # TODO(eeeqeee) add option to increment from current version
+      id: get-version
+      run: |
+        echo "version=${{ inputs.version }}" >> $GITHUB_OUTPUT
+    - name: Set git config
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
nguyentvan7, line 29: Where does this come from? is it just standard? You seem pretty safe with exposing info, but just wanna make sure it's not some repo-specific thing that leaks info somehow to have this visible publicly

.github/workflows/new-release.yml
```diff
@@ -0,0 +1,66 @@
...
+        git worktree add ../gh-pages gh-pages --no-checkout
+        cp -rT dist/apps/frontend ../gh-pages
+        cd ../gh-pages
+        git add --all .
+        git commit -m "deploy: version ${{ steps.get-version.outputs.version }} $(git rev-parse HEAD)"
+        git push
+    - name: Push
+      run: |
+        git status
+        git push
```
nguyentvan7, line 61: This pushes the version change to the main repo, I am guessing? This step will ignore the `cd ../gh-pages` from the previous step and still be in the correct directory?

nguyentvan7 (commented):
Mainly just questions, looks nice to me tho

.github/workflows/new-release.yml
eeeqeee: https://github.com/orgs/community/discussions/26560
Its the name+email for the Github Action bot

.github/workflows/new-release.yml
eeeqeee: This sets the env var `NX_URL_GITHUB_GO_CURRENT_VERSION` which should get picked up by `process.env`
It should link to the commit it was built with, since we already have the patch notes on the home page
If env var is unset (manual build) it should be like it was before

.github/workflows/new-release.yml
eeeqeee: yep, the current working directory should reset between steps

.github/workflows/new-release.yml
```diff
@@ -0,0 +1,66 @@
...
+          git add .
+          git commit -m "${{ steps.get-version.outputs.version }}"
+      - name: Yarn install
+        run: |
+          yarn install --immutable --immutable-cache
+      - name: Setup .env
+        run: |
+          echo $ENVFILE > apps/frontend/.env.local
+        env:
+          ENVFILE: ${{ secrets.ENVFILE }}
```
eeeqeee, line 44: i think the repo owner is the only one thats able to setup secrets, so we should wait to merge

.github/workflows/new-release.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+      - name: Change version
+        run: |
+          TEMP_FILE=$(mktemp)
+          jq --arg v "${{ steps.get-version.outputs.version }}" '.version = $v' package.json > $TEMP_FILE
+          cp $TEMP_FILE package.json
+          git add .
+          git commit -m "${{ steps.get-version.outputs.version }}"
```
nguyentvan7, line 36: If the version is unset, or the same as prior version, will this still deploy gracefully? Like if we want to redeploy with different secrets, without incrementing version

nguyentvan7: This commit should be empty, but not sure if the release code will fail silently or fail the whole action. Which isn't a huge deal since the deployment would have already happened, but still

.github/workflows/new-release.yml
nguyentvan7: Internet says we can manage secrets via rest API, but yea we can just wait for frzyc too, no rush

.github/workflows/new-release.yml
eeeqeee: It probably will fail to commit if its the same since the commit is empty, but I can change it
I can just add a toggle to avoid making a release which would probably be nicer than having part of the action fail

frzyc:
@eeeqeee can you provide some description on this is used, how the secrets are set, how to trigger action with specific version, etc...  eli5

eeeqeee:
So to setup secrets, you need to go to Settings > Secrets and variables > Actions > New repository secret and then create an new secret with name `ENVFILE` and copy the `.env.local` file into the Secret part
To run, you go to Actions > New Release > Run workflow and enter the new version number (or leave empty to rebuild without creating a new release) and click `Run Workflow`

nguyentvan7:
There should also be a way to setup github actions to allow any collaborator to upload secrets. afaik collaborators can edit repo secrets via REST api, but not via web interface. But for now, .env.local shouldn't change too often anyways so it should be ok to have only frzyc be able to update it

nguyentvan7 (approved):

frzyc (changes requested):

.github/workflows/new-release.yml
```diff
@@ -0,0 +1,75 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+      - name: Change version
+        if: steps.get-version.outputs.version != ''
+        run: |
+          TEMP_FILE=$(mktemp)
+          jq --arg v "${{ steps.get-version.outputs.version }}" '.version = $v' package.json > $TEMP_FILE
+          cp $TEMP_FILE package.json
+          git add .
+          git commit -m "${{ steps.get-version.outputs.version }}"
```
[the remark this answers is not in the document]
frzyc: also, i think you need to explicitly push the tag `git push origin "${{version}}"`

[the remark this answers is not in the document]
eeeqeee: `gh release create` should create the tag since it shouldn't already exist

frzyc (approved):

PR #1098, opened 2023-03-23
fix misc. sheet issues - Polomo 3.5 part 2 electric boogaloo

PolomoPT:
lmao such a van thing to say
so i revert xingqiu thingy from van's latest misc. ... and uh change mika strings and then in artifacts.json i deleted a few unused strings and fixed a couple typos thanks bye


PolomoPT:
basically uh yarn.lock is now back to how 'twas before
`what do you see when you revert a commit?`

nguyentvan7 (approved):

PR #1105, opened 2023-03-25
Avoid unnecessary rerender

### LINKED ISSUE #1001: Generating a build before all assets have loaded in gives incorrect results
If you change characters and click generate before the build page has fully loaded in the results are generated using the build settings of the previous character:

https://streamable.com/h3scvl

### LINKED ISSUE #1101: Extreme UI slowdown causing page rendering issues
Likely the same issue as #1001 but the UI slowdown is very apparent recently.

https://streamable.com/wa72yl

Although I could not recreate it in this video - sometimes opening the new character's multi-target popup before the page has fully rendered will overwrite their multi-targets with the previous character's.

eeeqeee:
Still suboptimal, but better than nothing
Fix #1101 
Fix #1001

`CharacterDisplayCard` was rerendering with an outdated `character`, use a key to get a new component

### DIFF
--- a/apps/frontend/src/app/PageCharacter/CharacterDisplay/index.tsx
+++ b/apps/frontend/src/app/PageCharacter/CharacterDisplay/index.tsx
@@ -82,7 +82,11 @@ export default function CharacterDisplay() {
         fallback={<Skeleton variant="rectangular" width="100%" height={1000} />}
       >
         {characterKey && (
-          <CharacterDisplayCard key={characterKey} characterKey={characterKey} onClose={onClose} />
+          <CharacterDisplayCard
+            key={characterKey}
+            characterKey={characterKey}
+            onClose={onClose}
+          />
         )}
       </Suspense>
     </Box>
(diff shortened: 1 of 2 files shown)


apps/frontend/src/app/PageCharacter/CharacterDisplay/index.tsx
```diff
@@ -82,7 +82,11 @@ export default function CharacterDisplay() {
         fallback={<Skeleton variant="rectangular" width="100%" height={1000} />}
       >
         {characterKey && (
-          <CharacterDisplayCard characterKey={characterKey} onClose={onClose} />
+          <CharacterDisplayCard
+            key={characterKey}
```
frzyc, line 86: why does assigning a `key` prevent rerender? curious.

eeeqeee: Currently we when we switch characters it first renders `CharacterDisplayCard` with the new `characterKey` and old `character` then it rerenders with the new `characterKey` and `character`. The time in between causes bugs
Now it unmounts the old `CharacterDisplayCard` and mounts a fresh copy which prevents this issue

frzyc (approved):

nguyentvan7 (approved):

Issue #1106, opened 2023-03-26
Show Conditional or Bonus stats separately on the character overview screen


StainAE86:
This:
![image](https://user-images.githubusercontent.com/101696552/227766197-21d8d088-45ae-4922-bd7a-895a1a6ebaae.png)

From this:
![image](https://user-images.githubusercontent.com/101696552/227765970-fd569611-401c-478b-b277-a2d76c56ccac.png)
![image](https://user-images.githubusercontent.com/101696552/227766230-34cc5840-8ed5-4579-9d53-fa54ed290810.png)
![image](https://user-images.githubusercontent.com/101696552/227766027-64c3c21a-9e20-462a-848f-7e86c29f6b1a.png)

nguyentvan7:
What is this asking for?

StainAE86:
The white number stats show the base character/weapon and artifact contribution only (as it would be in a "idle" situation in-game)

And the green +numbers show any conditional or bonus stats the user has configured in GO.

nguyentvan7:
Seems really confusing honestly. Especially since we only use the green/red numbers when doing comparisons. And this doesn't even follow the green number rule from in game, where arti stats are shown as green and base stats are in white

StainAE86:
Could show it in a different way. But it would be nice to show what the idle condition character looks like so its more of a 1:1 comparision between the game when you view your stats or Enka, etc.

StainAE86:
Similar suggestion from another user:
https://discord.com/channels/785153694478893126/1122395424075030658/1122395424075030658

PR #1111, opened 2023-04-01
Removes Conditionals from Wandering Evenstar, Xiphos Moonlight and Makhaira Aquamarine

PolomoPT:
I didn't mess with numbers or anything btw just thing about rendering and i checked whether it displayed correctly okay promise i didn't mess it up k.

I removed the duration thing too because the effect is continuous.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "genshin-optimizer",
-  "version": "9.6.3",
```
nguyentvan7, line 3: `git revert 90e8f49` or just change it bakc yourself

nguyentvan7 (commented):
I still don't know how I feel about this. I see the point in making it unconditional, but it still is kind of conditional, though only in cases where it makes no sense. Kinda related to #1106

nguyentvan7 (approved):

PR #1118, opened 2023-04-07
Update CharacterCard Header layout

frzyc:
Before
image

After
image

should be no top gap unless the layout is super cramped.

### DIFF
--- a/apps/frontend/src/app/Components/Character/CharacterCard.tsx
+++ b/apps/frontend/src/app/Components/Character/CharacterCard.tsx
@@ -34,9 +34,11 @@ import useTeamData from '../../ReactHooks/useTeamData'
 import type { ICachedArtifact } from '../../Types/artifact'
 import type { ICachedCharacter } from '../../Types/character'
 import { iconAsset } from '../../Util/AssetUtil'
+import type { RollColorKey } from '../../Types/consts'
 import { range } from '../../Util/Util'
 import ArtifactCardPico from '../Artifact/ArtifactCardPico'
 import CardLight from '../Card/CardLight'
+import ColorText from '../ColoredText'
 import ConditionalWrapper from '../ConditionalWrapper'
 import { NodeFieldDisplay } from '../FieldDisplay'
 import SqBadge from '../SqBadge'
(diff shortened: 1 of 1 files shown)


apps/frontend/src/app/Components/Character/CharacterCard.tsx
```diff
@@ -379,43 +373,43 @@ function HeaderContent() {
...
-          <Chip
-            color={burstBoost ? 'info' : 'secondary'}
-            label={<strong>{tBurst}</strong>}
-          />
-        </Grid>
-      </Grid>
-      <Typography mt={1}>
-        <StarsDisplay stars={characterSheet.rarity} colored />
+        </Box>
+        <Typography
```
nguyentvan7, line 385: Not a fan of this. maybe make the sqbadge the color? But having the text as the color looks tacky and makes it hard to read

apps/frontend/src/app/Components/Character/CharacterCard.tsx
```diff
@@ -379,43 +373,43 @@ function HeaderContent() {
...
+          label={<strong>{tSkill}</strong>}
+        />
+        <Chip
+          size="small"
+          color={burstBoost ? 'info' : 'secondary'}
+          label={<strong>{tBurst}</strong>}
+        />
+      </Box>
+      <Typography variant="h6" lineHeight={1}>
+        <StarsDisplay stars={characterSheet.rarity} colored inline />
```
nguyentvan7, line 412: Stars are small again

nguyentvan7 (changes requested):
koko is a little crushed, but I guess we can just hope there won't be any names longer than this for now 
image

apps/frontend/src/app/Components/Character/CharacterCard.tsx
frzyc: Stars are small again to make space, it was occupying a lot of vertical space.

nguyentvan7:
How much do you resize by? So I can do the same for future chars

apps/frontend/src/app/Components/Character/CharacterCard.tsx
frzyc: Sqbadge always looked weird because of the padding, would need to like half the font size and flex box vertical align, which probably looks weird...
image

nguyentvan7: Looks way better imo

nguyentvan7:
Pls stop force pushing it makes things more complex when reviewing and pulling

nguyentvan7 (approved):

frzyc:
its a  merge rebase, it rebases master's history below my changes, or else i will have to do resolve a lot of conflicts.

PR #1123, opened 2023-04-11
Fix Silly Optimizer implementation

nguyentvan7:
* Fixes incorrectly using silly wisher name when setting was disabled

I left a todo for UIData.tsx. I don't see an easy way to pass in the `silly` flag and it only affects the formula tooltip. If you see an easy way, let me know, I just didn't want to have to pass that flag everywhere. I could add it to the UIData class, but still seems kinda unnecessary, but we can do it, I think.


apps/frontend/src/app/PageDocumentation/index.tsx
```diff
@@ -346,7 +346,7 @@ function ArtifactSetKeyPane() {
   )
 }
 function CharacterKeyPane() {
-  const { t } = useTranslation('sillyWisher_charNames')
+  const { t } = useTranslation('charNames_gen')
```
frzyc, line 349:
```suggestion
  const { t } = useTranslation(['charNames_gen', 'sillyWisher_charNames'])
```

apps/frontend/src/app/Formula/uiData.tsx
```diff
@@ -559,7 +559,7 @@ function createDisplay(node: ContextNodeDisplay<number | string | undefined>) {
           <Translate ns="weaponNames_gen" key18={source} />
         )) ||
         (allCharacterSheetKeys.includes(source as CharacterSheetKey) && (
-          <Translate ns="sillyWisher_charNames" key18={source} />
+          <Translate ns="charNames_gen" key18={source} /> // todo: add silly support
```
frzyc, line 562: could use a new component that checks `sillyContext`

frzyc (approved):

Issue #1091, opened 2023-03-19
Extremely slow multi-target build time


StainAE86:
I'm not sure if this is due to the latest fix with the builder or its just a poorly created multi-target but its very slow.
Thought I'd bring it up since multi-targets are getting more popular and we've been promoting users leaving all artifact sets enabled as the default practice - and unless they drop the builds generated number down it will also default a higher number, 5 I think?

All sets, 10 builds: 3min 46s
![image](https://user-images.githubusercontent.com/101696552/226157195-1cd62a15-e211-44bc-ac6d-8c4fab1dd9ce.png)

All sets, 1 build: 52s
![image](https://user-images.githubusercontent.com/101696552/226157058-277b5abe-c07b-4815-85a6-0f5b6289f03d.png)

Emblem only, 1 build: 5s
![image](https://user-images.githubusercontent.com/101696552/226156968-904057be-39a6-4b0e-9588-8fd46321d374.png)

Full database:
[raiden multi.txt](https://github.com/frzyc/genshin-optimizer/files/11010375/raiden.multi.txt)

StainAE86:
The user who brought this to attention says they noticed it around the 13th of this month.

Lines up with this merge: https://github.com/frzyc/genshin-optimizer/pull/1004

tooflesswulf:
Ah, this is somewhat expected. The LP bounding is significantly more computationally costly than the old method; however it gives a much tighter bound so we can prune more aggressively. But to properly take advantage of more aggressive pruning, we need to implement a proper branching method. I'm about half-way done in https://github.com/frzyc/genshin-optimizer/tree/heuristic-branching. 

Aside from that, the bounding scales really badly in the number of dependent variables, which multi-opt for whatever reason adds a ton of. I noticed critrate & critdmg are duplicated like 10 times on the artifacts. I can also add a couple pre-processing steps to reduce the total number of variables (like merging flat atk and atk%).

PR #1133, opened 2023-04-14
Improve node equality detection

tooflesswulf:
Introduce a method to efficiently sort formulas, which also solves node duplication problems (most prevalent with large multi-opts).

Piggy-backs off #1131, together they seem somewhat effective at addressing #1091. On debug build, Raiden multi-opt compute time lowered from 75 to 15-ish seconds.


apps/frontend/src/app/Formula/optimization.test.ts
```diff
@@ -50,21 +47,42 @@ describe('optimization', () => {
...
...,"type":"number","value":0.18},{"operation":"const","operands":[],"type":"number","value":0}],"emptyOn":"l"},{"operation":"min","operands":[{"operation":"const","operands":[],"type":"number","value":0.8},{"operation":"mul","operands":[{"operation":"const","operands":[],"type":"number","value":0.28},{"operation":"add","operands":[{"operation":"const","operands":[],"type":"number","value":-1},{"operation":"add","operands":[{"operation":"read","operands":[],"path":["dyn","0"],"accu":"add","type":"number"},{"operation":"threshold","operands":[{"operation":"read","operands":[],"path":["dyn","EmblemOfSeveredFate"],"accu":"add","type":"number"},{"operation":"const","operands":[],"type":"number","value":2},{"operation":"const","operands":[],"type":"number","value":0.2},{"operation":"const","operands":[],"type":"number","value":0}],"emptyOn":"l"},{"operation":"threshold","operands":[{"operation":"read","operands":[],"path":["dyn","TheExile"],"accu":"add","type":"number"},{"operation":"const","operands":[],"type":"number","value":2},{"operation":"const","operands":[],"type":"number","value":0.2},{"operation":"const","operands":[],"type":"number","value":0}],"emptyOn":"l"}]}]}]}]}]}]}]}]}]}]}
```
lantua, line 53: @frzyc not sure if we should separate this into a new json. thoughts?

lantua (changes requested):

apps/frontend/src/app/Formula/optimization.ts
```diff
@@ -357,7 +378,7 @@ export function constantFold(
 
           if (numericValue !== f([]))
             // Skip vacuous values
-            formulaOperands.push(constant(numericValue))
+            formulaOperands.unshift(constant(numericValue))
```
lantua, line 381: If it will be reordered anyway, we shouldn't need `unshift`, which is normally O(n).

apps/frontend/src/app/Formula/optimization.test.ts
```diff
@@ -50,21 +47,42 @@ describe('optimization', () => {
...
-    const f1 = sum(r1, r2, r3, r4, r5, r6),
-      f2 = sum(r3, r4)
-    // TODO: Factoring process may reorder the term ( r3 + r4 ). May need to update the checking
-    expect(deduplicate([f1, f2])).toEqual([
-      sum(r1, r2, r5, r6, sum(r3, r4)),
-      sum(r3, r4),
-    ])
+  test('complex formula invariant', () => {
+    // Raiden N3C1DD or something idk
+    // prettier-ignore
```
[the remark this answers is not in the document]
tooflesswulf: Actually I should take this particular test case out b/c constantFold & deduplicate are no longer stationary with each other. But for the future a test set of large formulas & fake artifacts could be very useful.

lantua (approved):

PR #1138, opened 2023-04-15
Improve formula reduction (reaffine)

tooflesswulf:
Now breaks apart sums of affine & non-affine expressions, then combines all affine sub-expressions into a single new variable.

Example `ATK + 1500*ATK% + 1500*thresh(Glad, 2, 18%, 0)` -> `r0 + 1500*thresh(Glad, 2, 18%, 0)`


nguyentvan7 (approved):
Approving assuming it's the same as #1138

PR #1139, opened 2023-04-15
Improve formula reduction (deduplication)

tooflesswulf:
Recognizes identical nodes and merges them into the same reference, helping the solver avoid unecessary re-calculations.


nguyentvan7 (approved):
Approving assuming it's the same as #1133

PR #1140, opened 2023-04-15
Add transformative reaction defense multiplier

nguyentvan7:
image
image
image

### DIFF
--- a/apps/frontend/src/app/Formula/index.ts
+++ b/apps/frontend/src/app/Formula/index.ts
@@ -274,6 +274,7 @@ const input = setReadNodeKeys(
 
     enemy: {
       def: read('add', { ...KeyMap.info('enemyDef_multi_'), pivot }),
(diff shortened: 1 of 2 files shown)


frzyc:
Can you add some context picture of how this will show up in formula tooltip/detailed formula?

apps/frontend/src/app/Formula/index.ts
```diff
@@ -532,6 +533,14 @@ const common: Data = {
         sum(one, prod(-1, enemy.defIgn))
       )
     ),
+    transDef: frac(
+      99999999,
+      prod(
```
frzyc, line 538: Shouldn't this be using the calculated `def` instead of being calculating again?

nguyentvan7: I don't see any mathematical way to use the other def term in here

lantua (approved):

PR #1149, opened 2023-04-21
Fix graph showing 00 when there are no builds

nguyentvan7:


frzyc:
Can you provide a before/after? you seem to fix one `0` by converting it to a boolean, but where is the other `0`?

nguyentvan7:
Think the 2 0s showed up due to JavaScript trying to cast everything together. Most likely something was just an empty array, which when cast together with length turned into 00. Which is why everything should be booleanized for conditional rendering

frzyc (approved):

PR #1150, opened 2023-04-21
Fix exclusion list enabling when editing the list

### LINKED ISSUE #1088: Adding an artifact to the "Exclude" list enables using artifacts from the exclude list
When you add (or remove) an artifact manually via the exclude artifact popup the list gets enabled causing the builder to include what you've just specifically setup to exclude.

image

image

nguyentvan7:
* Resolve #1088


frzyc (changes requested):
~I have a question in #1088 that needs to be clarified.~
I have a change in https://github.com/frzyc/genshin-optimizer/pull/1153 that expands on this.

Issue #1155, opened 2023-04-23
Builder crash


StainAE86:
If you generate for Nilou in this database using the current settings the builder will crash.

Removing teammates/constellations/refinements doesn't fix it.
Changing the hitmode from Vape back to No Reactions or utilizing the Artifact Set Config to disable everything not suitable for Nilou will allow the builder to successfully finish.

[generate for Nilou.txt](https://github.com/frzyc/genshin-optimizer/files/11303160/generate.for.Nilou.txt)

tooflesswulf:
Im dead 💀 💀 

Okay so this highlights so many things. First, the crashing is caused by excessive memory consumption. That part is already addressed by #1139 though.

Second, build times are monumentally slow because the polynomial representation of the formula (when expanded) takes like 2,500 terms. And half of that is caused by the `Vourukasha's Glow` artifact set b/c C6 nilou ult is 4th order in that set. (it gives hp, critRate, critDmg, and dmgBonus). There is no bug in the implementation as far as I can tell.

To address this, I'll have to improve the `expandPoly` function to recognize that (for example) 4P Emblem & 2P TotM cannot both contribute to the damage at the same time.

nguyentvan7:
![Mimee_2023-11-21_17-16-08.txt](https://github.com/frzyc/genshin-optimizer/files/13429332/Mimee_2023-11-21_17-16-08.txt)

Db doesn't have enough vglow pieces but still chokes on c6 nilou

PR #1156, opened 2023-04-24
Ignore Cross-Terms Between Set Effects

tooflesswulf:
An example of cross-terms between sets in an optimization formula is:
```
( atk + atk% + thresh(Glad, 2) + thresh(Echoes, 4) ) * ( EM + thresh(Gilded, 2) + thresh(Gilded, 4) )
```

When expanding, obviously the `thresh(Echoes, 4) * thresh(Gilded, 2)` product is always zero because we only have 4 artifact slots. This PR addresses such cross-terms the the polynomial expansion.

Issue was revealed by #1155, but don't resolve that one just yet because It's not guaranteed this PR fixes it.


apps/frontend/src/app/Solver/GOSolver/BNBSplitWorker/polyUB.ts
```diff
@@ -303,71 +319,89 @@ export type SumOfMonomials = Monomial[]
...
+                nxt.setUsage[k]
+              ))
+          )
+          return ret
+        },
+        { $k: 1, terms: [], setUsage: {} }
+      )
+    )
+    .filter(
+      ({ setUsage }) => Object.values(setUsage).reduce((a, b) => a + b, 0) < 5
```
eeeqeee, line 354: We have 5 artifact slots, so shouldn't be
```suggestion
      ({ setUsage }) => Object.values(setUsage).reduce((a, b) => a + b, 0) <= 5
```

nguyentvan7:
Benefits were not tangible

PR #1159, opened 2023-04-24
Fix optimizer artifact settings text

### LINKED ISSUE #1154: // in "Rainbow Builds" title
image
image

vapopescu:
Fixes #1154
Other languages also tested and working

image


nguyentvan7 (approved):

frzyc:
@PolomoPT I think we need to update POEditor with the new strings changed here, so when we next pull from POeditor it won't overwrite this. This way we can retain the existing translation(since the change only changes the numbers in the string)

PR #1160, opened 2023-04-24
Compute Raiden burst energy generation dynamically

vapopescu:
- Dynamically calculate the value for the energy restoration effect of the Musou Isshin attacks
- Change the description (English only) of the A4 passive

Result:
Screenshot 2023-04-22 231804


apps/frontend/src/app/Data/Characters/RaidenShogun/index.tsx
```diff
@@ -228,6 +238,10 @@ const dmgFormulas = {
     plunge: burstResolve(dm.burst.plunge),
     plungeLow: burstResolve(dm.burst.plungeLow),
     plungeHigh: burstResolve(dm.burst.plungeHigh),
+    energyGen: prod(
+      subscript(input.total.burstIndex, dm.burst.enerGen),
+      sum(1, a4EnergyRestore_)
```
nguyentvan7, line 243:
```suggestion
      sum(one, a4EnergyRestore_)
```
You'll also need to update the import

nguyentvan7 (changes requested):

apps/frontend/src/app/Data/Characters/RaidenShogun/index.tsx
vapopescu: Tooltip now showing percentage, as expected.

image

nguyentvan7 (approved):

PR #1169, opened 2023-04-28
changes Baizhu and Kaveh strings

PolomoPT:
a little more with like how conds display but no biggie.

### DIFF
--- a/apps/frontend/src/app/Data/Characters/Kaveh/index.tsx
+++ b/apps/frontend/src/app/Data/Characters/Kaveh/index.tsx
@@ -414,7 +414,6 @@ const sheet: ICharacterSheet = {
     ]),
     constellation6: ct.talentTem('constellation6', [
       ct.fieldsTem('constellation6', {
-        canShow: equal(condAfterBurst, 'on', 1),
         fields: [
           {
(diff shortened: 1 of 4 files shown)


apps/frontend/src/app/Data/Characters/Kaveh/index.tsx
```diff
@@ -427,8 +412,24 @@ const sheet: ICharacterSheet = {
...
       { fields: [{ node: skillC5 }] },
     ]),
-    constellation6: ct.talentTem('constellation6'),
-  },
+    constellation6: ct.talentTem('constellation6', [
+      ct.fieldsTem('constellation6', {
+        canShow: equal(condAfterBurst, 'on', 1),
+        fields: [
+          {
+            node: infoMut(dmgFormulas.constellation6.dmg, {
```
nguyentvan7, line 420: actually on second thought, this shouldn't be behind a conditional at all

PolomoPT: yeah i thought that too but since it used to be... will change

nguyentvan7 (approved):

PR #1170, opened 2023-04-28
Fix parsing slowness

### LINKED ISSUE #1108: Linting for `waverider` is slow
Linting for `waverider` is slower and uses more memory than the frontend and other libraries

nguyentvan7:
* Resolve #1108 

I'm still not sure on the specifics, but there must be some large dependency or something inside the `waverider` module that was causing the slowness. Maybe some reliance on `typescript` module or something that is large, according to Google. Either way, we were getting hit by [this bug](https://github.com/typescript-eslint/typescript-eslint/issues/1192#issuecomment-552990973) because of our parser setup. This caused slowness and memory usage issues. I did the fix in that thread a few comments down and it is significantly faster to parse and lint `waverider` module, and as a result, this also fixes slowness in the `formula` module

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,11 +25,7 @@
       "files": ["*.ts", "*.tsx"],
       "extends": ["plugin:@nrwl/nx/typescript"],
       "parserOptions": {
-        "project": [
-          "apps/*/tsconfig.json",
-          "apps/*/tsconfig.*?.json",
-          "libs/*/tsconfig.*?.json"
-        ]
+        "project": "./tsconfig.eslint.json"
       },
       "rules": {
         "@typescript-eslint/consistent-type-exports": "warn",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           yarn install --immutable --immutable-cache
       - name: Run lint
         run: |
-          NODE_OPTIONS="--max-old-space-size=4096" yarn run nx run-many --target=lint --max-warnings=0
+          yarn run nx run-many --target=lint --max-warnings=0
   test:
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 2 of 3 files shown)


nguyentvan7:
You can get timings for specific eslint rules by setting an environment variable `TIMING=1`, then running `npx eslint <files>`.

I tested this way on a `formula` module file, but all rules ran in < 5 ms, so there was no issue with rules. [Some comment on Stack Overflow](https://stackoverflow.com/questions/38458067/which-eslint-rules-in-my-config-are-slow) led me to believe it was related to parsing.

image

You can get timings for parsing/all of eslint by setting an environment variable `DEBUG=eslint:*`. This produces a lot of fluff, and once I narrowed it down to parsing/linting, I set the env varable `DEBUG=eslint:linter`.

Running this, it was taking 7s to parse Nilou's entire 600-line sheet. Whereas `formula` file of 61 lines took 28s
image
image

Looking up stuff related to parsing slowness in monorepo, I ended up at the link in the description. Making the changes in this PR changed Nilou to 10s parse, and `formula` file to 9s parse. There was also clearly noticeable speedups in linting within vscode for `formula` module. Before, I had to wait 30s-1 minute for lint errors to show up, and now they are showing instantly, similar to our other modules.

nguyentvan7:
Linting time for another PR was 9 minutes https://github.com/frzyc/genshin-optimizer/actions/runs/4833269220

Linting time for this PR was 2.5 minutes https://github.com/frzyc/genshin-optimizer/actions/runs/4834551217

nguyentvan7:
Just as a tldr, before, if we had some dependencies between modules, we would load each dependency everytime, very redundantly. So if we had a dependency like

* A depends on C
* b depends on C

We would actually have to generate the typing/ast for C and have 3 copies of it in memory. Once for A's dependency, once for B's dependency and once for C itself

The change makes it so our project is generated as a whole for linting and loaded entirely in memory. So it should be okay as long as our project is not humongous, though I don't know what qualifies as humongous

eeeqeee (approved):

nguyentvan7:
Waiting to see if nx 16 fixes or breaks this

eeeqeee (approved):

PR #1175, opened 2023-04-29
Adjust maxIterateSize

tooflesswulf:
Increase enumerate threshold from 4M to 32M. Up to 2x speedup depending on problem on my local machine.

Currently the runtime bottleneck is the splitting process. Increasing the enumeration simultaneously 1. increases the work of the enumeration workers and 2. decreases the level that the splitter needs to go to. In effect, it better balances the time spent between splitting and enumerating. However, the optimal value for this constant will be system-dependent & thread-count dependent.

### DIFF
--- a/apps/frontend/src/app/Solver/GOSolver/GOSolver.ts
+++ b/apps/frontend/src/app/Solver/GOSolver/GOSolver.ts
@@ -12,7 +12,7 @@ import { pruneAll, pruneExclusion } from '../common'
 import { WorkerCoordinator } from '../coordinator'
 
 export class GOSolver extends WorkerCoordinator<WorkerCommand, WorkerResult> {
-  private maxIterateSize = 4_000_000
+  private maxIterateSize = 32_000_000
   private status: Record<'tested' | 'failed' | 'skipped' | 'total', number>
   private exclusion: Count['exclusion']
   private topN: number


lantua (approved):

nguyentvan7:
Is there any heuristic we use to determine the maxIterateSize or are we just trial-and-erroring until we get a good value?

tooflesswulf:
We're just trial-and-error checking different sizes. 32M happens to work well on my machine & lantua's I assume.

lantua:
> Is there any heuristic we use to determine the maxIterateSize or are we just trial-and-erroring until we get a good value?

There are two trivial restrictions;
- Small `maxIterateSize` puts stress on `SplitWorker`, and `ComputeWorker` has to wait for each splitting step, essentially serializing the process, and
- Large `maxIterateSize` leads to many builds being unnecessarily checked when they could be pruned by the `SplitWorker`.

It still largely depends on env, but those restrictions put `maxIterateSize` somewhere between 10–50M before I experience a sharp drop in performance.
I'm not 100% sure how we could home in on the exact value beyond that, though.
Also, this range could change with the performance profiles of `SplitWorker` and `ComputeWorker`, so we need to retune this value once in a while, e.g., when the splitting heuristic changes.

lantua:
It also depends on the optimization problem, so there's a risk of overfitting if the test cases are not diverse enough.

PR #1176, opened 2023-05-02
Worker work sharing and yielding

tooflesswulf:
Somewhat elegant inter-worker communication methods. No significant changes in runtime observed, but the solver execution order is totally changed. Feels slightly better for dopamine though :p (less big jumps 0-100%, but choppier in the middle)

The idea was to (A.) share the splitting task between the workers so the bottleneck is not entirely on 1 splitter. However, because splitting "blocks" until it is completely done, we need to add (B.) a preempting process to force the splitters to pause and enumerate a few iterate commands when they overflow.

Implementation-wise, 3 main components.
1. `notifyEmpty()` - Idea A - When the coordinator sees 0 commands left in queue & there may be idle workers, the coordinator will send a `ShareRequest` to all workers to insert a couple `split` commands for the idle workers.
2. `notifyCommandOverflow()` - Idea B - When the coordinator notices excess iterate commands, it'll pop a few off and distribute them to the workers. The workDone() trackers are updated to track that it needs to receive "done" twice.
3. `BackgroundWorker::manualEventLoop` - When a worker receives a command, it stores the command in a queue so they must be resolved before the next split() call. I manually queue these commands instead of using the default js event loop due to below issues.

==========

Issues I encountered with event loop stuff:
- Default js event loop doesn't seem to like switching between the threads, so it stays on the split() task for a while, even when there are 50-100 iterate() tasks waiting for their turn. This causes the pending iterate tasks to build up inside the worker.
- Default js event loop seems to struggle with scheduling when there are hundreds of pending iterate() commands. Not sure why, or if I did something wrong in implementation. This lead to massive browser slowdown, to the point the tab no longer even responds.


frzyc:
A few things after cursory glance at the code and your desc:
1. Javascript is basically single threaded, that means that if you have something running in its event loop, nothing else can interrupt it. You probably hit the limit to what you can do in terms of splitting/managing on the front-end/manager thread.
2. In an ideal situation, we would have a main worker that splits work against its sub workers, this would probably make our situation a lot easier. The historical reason why we have all this worker management on the frontend is that when I initially implemented multithreading, Safari(WebKit) was unable to have nested dedicated workers(was a lot of frantic recoding to move all the manager code onto the frontend)
3. However, I don't believe 2) to still be the case. I believe that in [2022 September, nested dedicated workers are actually part of WebKit](https://github.com/WebKit/WebKit/commit/48880e342359f100878b1b87373e706db7dfe540). This might mean that if we can assume enough adoption, perhaps we can actually implement workers as originally intended.(cope)

Through for food.

nguyentvan7:
Closed because improvements were marginal at best and complexity/issues seemed not worth it

PR #1183, opened 2023-05-23
Add 3.7 content

nguyentvan7:
* Add Kirara
* Add Ibis Piercer
* Fix Baizhu silly wisher name


nguyentvan7:
@PolomoPT please review strings

libs/localization/assets/locales/en/char_Kirara.json
```diff
@@ -0,0 +1,6 @@
+{
+  "maxShield": "Max DMG Absorption",
```
frzyc, line 2: Is this a common string?

littlef0x: It's not common because there are only two characters who can stack shield. But the item should be correct. See [Thoma](https://github.com/frzyc/genshin-optimizer/blob/master/libs/localization/assets/locales/en/char_Thoma.json) (except that's maximum but not max).
```
  "maxShield": "Maximum DMG Absorption",
  "maxPyroShield": "Maximum Pyro DMG Absorption"
```

The in-game text is 'Max Shield DMG Absorption'. So if you want to be as accurate as possible, both should be changed.

frzyc (approved):

PR #1189, opened 2023-05-28
Fix misc. sheet issues

### LINKED ISSUE #1180: Yelan NA should be visible even in C6
This allows for accurate multi opt during C6. Also mastermind state DMG shouldn't be hidden behind a conditional

### LINKED ISSUE #1181: Baizhu C2 healing is probably incorrect
The 20% multiplier is only applied to the percentage value and not the flat value of the heal. That is probably wrong, tho I guess it would be best for someone to verify

### LINKED ISSUE #1187: Kirara C1 is not calculated correctly

nguyentvan7:
* Fix #1187 
* Fix #1181
* Fix #1180 
* Fix Baizhu region
* Fix Baizhu silly name frfr this time

### DIFF
--- a/apps/frontend/src/app/Data/Characters/Baizhu/index.tsx
+++ b/apps/frontend/src/app/Data/Characters/Baizhu/index.tsx
@@ -206,7 +206,10 @@ const dmgFormulas = {
           subscript(input.total.skillIndex, dm.skill.healHp, { unit: '%' }),
           percent(dm.constellation2.heal)
         ),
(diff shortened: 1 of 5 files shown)


frzyc:
Do we need a more formalized system to check against lacking the silly name? I assume this will happen for future char releases

nguyentvan7:
We can use exists function of i18n. Looks easy to add lemme test it

nguyentvan7:
Sounds like a separate pr

frzyc (approved):

PR #1190, opened 2023-05-28
Add 2p/4p toggle in artifact comparison modal

### LINKED ISSUE #1151: Allow artifacts 2p/4p sets to be disabled directly from the build results
Implement the final idea: https://github.com/frzyc/genshin-optimizer/issues/1151#issuecomment-1518389725

**Origional Issue:** This should be default behaviour only if "all" artifacts are allowed on that character. However if you have a character with a custom list (eg; vv or bs character for example) you now have to disable the 2p and 4p of the new sets.

It would be nice for new sets to be off by default if the artifact set config is not in the default "all on" state.

image
image

nguyentvan7:
* Resolve #1151 

image

### DIFF
--- a/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/SetInclusionButton.tsx
+++ b/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/SetInclusionButton.tsx
@@ -12,10 +12,10 @@ import useBuildSetting from '../useBuildSetting'
 
 export default function SetInclusionButton({
   setKey,
-  sx,
+  buttonGroupSx,
 }: {
   setKey: ArtifactSetKey
-  sx?: SxProps<Theme>
+  buttonGroupSx?: SxProps<Theme>
 }) {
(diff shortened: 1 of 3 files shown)


apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/SetInclusionButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Button, ButtonGroup } from '@mui/material'
+import { useContext } from 'react'
+import { useTranslation } from 'react-i18next'
+import { CharacterContext } from '../../../../../Context/CharacterContext'
+import { handleArtSetExclusion } from '../../../../../Database/DataManagers/BuildSettingData'
+import useBuildSetting from '../useBuildSetting'
+
+export default function SetInclusionButton({
+  setKey,
+  sx,
```
frzyc, line 15: nit: recommend renaming to `buttonGroupSx`

frzyc (approved):

PR #1197, opened 2023-06-08
migrate to nx 15.9.4

frzyc:
This is the most minimal migration before nx v16.


eeeqeee (commented):
Is there a specific reason we need to upgrade? nx 16 is out so we might want to look into that

frzyc:
> Is there a specific reason we need to upgrade? nx 16 is out so we might want to look into that

Yeah, I've had some attempts to migrate to 16.3, but there is currently a bug with the latest release that breaks the js:node executors. Once that is addressed, we should be able to migrate to 16.4.
when I attempted the migration, I had some problems with migrating directly to 16.3, but if i migrate partially to 15.9.4, then migrate to 16.3, it was a much smoother experience. 

After some digging, I realized that the full migration failed because of this single line in package.json:
```
"nx": "^15.7.2",
```
After removing the `^`, which is removed in this PR as well, migration to 16 should be smooth.

PR #1204, opened 2023-06-15
add typing to any functions in util

frzyc:
Mainly fix the `any`s in `@genshin-optimizer/util`/`object.ts`


apps/frontend/src/app/Formula/api.tsx
```diff
@@ -303,7 +303,7 @@ function uiDataForTeam(
     crawlObject(
       buff ?? {},
       [],
-      (x) => x.operation,
+      (x) => (x as any).operation,
```
frzyc, line 306: @lantua I am not sure how to fix this type, suggestions?

lantua: You mean removing explicit `any`? we probably need to refactor `Data` type hierarchy or cast to `NumNode | StrNode | { operation?: undefined }`, which I don't think make things much clearer.

libs/util/src/lib/object.ts
```diff
@@ -13,17 +14,20 @@ export function crawlObject(
...
+  obj: Obj,
   keys: readonly (number | string)[],
   value: T
-) {
+): Obj {
   keys.reduce((accu, key, i, arr) => {
-    if (i === arr.length - 1) return (accu[key] = value)
+    if (i === arr.length - 1) {
+      accu[key] = value
+      return accu
```
lantua, line 26: This is a different return value, did you mean `accu[key]`, or has the semantic changed?

frzyc: the return value here doesn't matter, because this is the last iteration of the `reduce` function, and we are not using the returned value of the `reduce` function. returning `accu` makes sure that at all paths, the same shape is being returned back(even in a logic branch that doesn't matter"

PR #1209, opened 2023-06-24
Artifact Upgrade Probability Calculator (dev-mode only)

tooflesswulf:
The math is wrapped inside a class nicely now. Should be fairly straightforward:
1. Construct a UpOptCalculator class w/ some nodes, thresholds, and current build.
2. Add each artifact to the calculator one by one
3. Perform loose (fast) calculations with calcFastAll()
4. Perform more accurate (slow) calculations on top few candidates

```js
upopt = new UpOptCalculator(
    nodes,
    thresholds,
    curEquip
)
artifacts.forEach(art => upopt.addArtifact(art))
upopt.calcFastAll()
upopt.calcSlowToIndex(5)  // Top 5 artifact recommendations stored in upopt.artifacts[0:5]
```


frzyc:
Convert to "Ready to review" just to check deployment

lantua:
Based on #1237, we have three types of calculations: slow, fast, and exact.
We use the slow and fast calculations for the ranking, but the exact one to display to the users.
There seems to be no bounded relationship between the three calculations, albeit in practice, they tend to approximate one another.
This may confuse if the ranking and the displayed numbers do not agree (e.g. if the first result has a smaller value than the third one).

We may need to put a warning to the users, and/or perform the exact calculation in the background and update the ranking as the result becomes available.

nguyentvan7:
Validated that the upopt page is not present in UI and will not load if devmode is disabled

frzyc (approved):

apps/frontend/src/app/Components/NoArtWarning.tsx
```diff
@@ -0,0 +1,15 @@
+import { Alert, Link } from '@mui/material'
+import { Link as RouterLink } from 'react-router-dom'
+
+export default function NoArtWarning() {
+  return (
+    <Alert severity="warning" variant="filled">
```
nguyentvan7, line 6: We have a translation key for this one, we should add it to this shared component

PR #1211, opened 2023-06-26
Add basic character card and stat display functionality

nguyentvan7:


apps/gi-frontend/src/app/Components/Character/CharacterCard.tsx
```diff
@@ -0,0 +1,607 @@
...
+    <Suspense
+      fallback={
+        <Skeleton
+          variant="rectangular"
+          width="100%"
+          height={hideStats ? 300 : 600}
+        />
+      }
+    >
+      {/* <CardLight
```
frzyc, line 189: you can use `CardThemed` from `@genshin-optimizer/ui-common`
```
<CardThemed bgt="light">
```

PR #1224, opened 2023-07-02
Migrate gi-frontend-web to gi-frontend on vite with new char card

nguyentvan7:
image


frzyc:
Should also remove `gi-frontend-web`

libs/pando/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "module": "commonjs",
+    "module": "ES2022",
```
nguyentvan7, line 4: @lantua is this bad

lantua: if it compiles (and test succeed), should be fine

frzyc (approved):
Good enough, 🚀

PR #1225, opened 2023-07-03
Force vsCode to use prettier

frzyc:


.vscode/settings.json
```diff
@@ -6,5 +6,12 @@
   "files.trimTrailingWhitespace": true,
   "typescript.preferences.importModuleSpecifier": "project-relative",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "files.eol": "\n"
+  "files.eol": "\n",
+  "prettier.configPath": ".prettierrc",
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
nguyentvan7, line 12: This really shouldn't be necessary. It's most likely because you have user specified settings which overrides workspace

PR #1226, opened 2023-07-04
Fix effects active w/o fulfilled conditions, correct Jean's and Amber's C2 and C1 opt nodes, KQM Multi-Opt link.

PolomoPT:
Effects being active when they shouldn't (constellation or ascension requirements not met)

Plus I removed some opt nodes from Jean because they're not scaling. **And I changed Amber's things in a way that breaks multi-opts but is needed**
And I'm taking the opportunity to update the KQM link thing I wanted to do.


apps/frontend/src/app/Data/Characters/Yoimiya/index.tsx
```diff
@@ -144,7 +144,11 @@ const atk_ = greaterEq(
     )
   )
 )
-const c1atk_ = equal(condC1, 'c1', percent(dm.constellation1.atk_))
+const c1atk_ = greaterEq(
```
frzyc, line 147: @nguyentvan7 just based on intuition, `condC1` should be locked behind the constellation check as well, right?

libs/localization/assets/locales/en/char_AratakiItto.json
```diff
@@ -1,7 +1,4 @@
 {
-  "a1": {
-    "name": "Consecutive Arataki Kesagiri uses"
-  },
```
eeeqeee, line 4: This is still in use 
https://github.com/frzyc/genshin-optimizer/blob/97b5208f8b36d5958e77da87631327965d60b4fb/apps/frontend/src/app/Data/Characters/AratakiItto/index.tsx#L274

eeeqeee (changes requested):

libs/localization/assets/locales/en/char_AratakiItto.json
PolomoPT: oh i reverted something i did but forgot to revert that ig oof

nguyentvan7 (approved):

apps/frontend/.env
```diff
@@ -11,7 +11,7 @@ NX_URL_GITHUB_VAN=https://github.com/
 NX_GA_TRACKINGID=UA-000000-01
 NX_URL_YOUTUBE_TUTPL=https://www.youtube.com/
 NX_URL_WEBSITE_KQM=https://keqingmains.com/
-NX_URL_KQM_MULTI_GUIDE=https://docs.google.com/document/d/12hrPJ3MrAF7t03gyiiVtV5uI_rUKlVlbEwl9Lf2esIQ/edit?usp=sharing
+NX_URL_KQM_MULTI_GUIDE=https://keqingmains.com/misc/multi-optimization/
```
nguyentvan7, line 14: @frzyc we need to change the envfile repo secret. I've updated it on the drive, please update the repo value to match, thanks

also frzyc, please resolve this comment when done, then we can merge the pr

nguyentvan7 (approved):

apps/frontend/.env
eeeqeee: If we leave `NX_URL_KQM_MULTI_GUIDE` unset in the repo secret it should just pull from this file

PR #1229, opened 2023-07-09
Update `util` module

lantua:
Currently, `util` module has less restriction than some other modules, so the those module try to include `util`, the compilation will fail. This add more restriction to `ts` configuration.


frzyc (approved):

lantua:
Superseded by #1231

PR #1231, opened 2023-07-11
House Cleaning

lantua:
Several overall cleaning
- Move nx target's configuration to `nx.json` for more uniform behaviour
- Fix a bug where `build` targets use path relative to current directory, now default to workspace root
- Refactor common tsconfig to `tsconfig.base.json`, alleviating project fragility where type-checking may fail when one project imports another project
- Move several `pipeline` scripts into dedicated executors


package.json
```diff
@@ -6,23 +6,26 @@
...
+    "mini-ci": "nx affected -t test,lint --max-warnings=0"
   },
   "devDependencies": {
     "@babel/preset-react": "^7.18.6",
     "@emotion/babel-plugin": "11.11.0",
     "@nx/cypress": "16.4.3",
     "@nx/eslint-plugin": "16.4.3",
-    "@nx/jest": "16.4.3",
-    "@nx/js": "16.4.3",
+    "@nx/jest": "16.5.1",
```
frzyc, line 17: im not sure about bumping nx module version without bumping `nx`. i would recommend running `nx migrate latest` to keep them in sync.

apps/frontend/project.json
```diff
@@ -92,20 +93,8 @@
       "command": "gh-pages -d dist/apps/frontend",
```
frzyc, line 93: let's remove the deploy command, since we are using CI to deploy, manual deploy can actually break https://github.com/frzyc/genshin-optimizer/pull/1163 in the foreseeable future

.vscode/settings.json
```diff
@@ -6,5 +6,6 @@
   "files.trimTrailingWhitespace": true,
   "typescript.preferences.importModuleSpecifier": "project-relative",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "files.eol": "\n"
+  "files.eol": "\n",
+  "eslint.validate": ["json"]
```
eeeqeee, line 10: This is legacy according to https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint .
```suggestion
  "files.eol": "\n"
```

package.json
```diff
@@ -6,23 +6,26 @@
...
     "@nx/linter": "16.4.3",
+    "@nx/plugin": "^16.5.1",
     "@nx/react": "16.4.3",
     "@nx/vite": "16.4.3",
     "@nx/web": "16.4.3",
     "@nx/webpack": "16.4.3",
     "@nx/workspace": "16.4.3",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.10",
     "@svgr/webpack": "8.0.1",
+    "@swc-node/register": "~1.4.2",
```
eeeqeee, line 28: What is this used for?

libs/plugin/src/executors/sync-repo/executor.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const options: SyncRepoExecutorSchema = {
+  repoUrl: '',
+  localPath: '',
+}
+
+describe.skip('SyncRepo Executor', () => {
+  it('can run', async () => {
+    const output = await executor(options)
+    expect(output.success).toBe(true)
```
frzyc, line 12: does this actually test anything? i would almost expect it to fail on empty strings...

lantua: It's the default, and skipped rn. Any idea what to test on something that fetch a repo?

package.json
lantua: Nx seems to be using this to run executor (sync-repo)

package.json
lantua: That wasn't the intension, `yarn` seems to bump it automatically when I add `@nx/plugin`

package.json
lantua: We can remove it and switch to the "slower" tsc

libs/plugin/src/executors/sync-repo/executor.ts
```diff
@@ -11,18 +11,20 @@ export default async function runExecutor(
   const name = path.basename(cwd)
 
   // Fetch existing or clone a new repo
-  if (fs.existsSync(cwd)) execSync('git fetch -q --depth 1', { cwd })
```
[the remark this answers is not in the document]
lantua: ?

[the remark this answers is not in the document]
frzyc: why `why`?

libs/plugin/src/executors/sync-repo/executor.spec.ts
frzyc: we could fetch a repo that will never get removed from github. I recommend https://github.com/chrislgarry/Apollo-11/

libs/plugin/src/executors/sync-repo/executor.ts
```diff
@@ -0,0 +1,43 @@
...
+  const name = path.basename(cwd)
+
+  if (fs.existsSync(cwd)) {
+    // Fetch & reset
+    const remoteHash = getRemoteRepoHash(url)
+    const localHash = getLocalRepoHash(cwd)
+    if (remoteHash !== localHash) {
+      console.log('Local and remote hashes differ, fetching')
+      execSync(`git fetch -q --depth 1`, { cwd })
+      execSync(`git reset -q --hard origin/master`, { cwd })
```
eeeqeee, line 21: Do we need these to be quiet? it would make it harder to debug on failure and it should rarely run anyway so it shouldnt spam the console
```suggestion
      execSync(`git fetch --depth 1`, { cwd })
      execSync(`git reset --hard origin/master`, { cwd })
```

libs/plugin/src/executors/sync-repo/executor.ts
```diff
@@ -0,0 +1,43 @@
...
+    } else console.log('Repo already existed with the latest commit')
+  } else {
+    // Clone
+    console.log(`Cloning ${name}`)
+    const parent = path.dirname(cwd)
+    fs.mkdirSync(parent, { recursive: true })
+    execSync(`git clone ${url} -q --depth 1 ${name}`, { cwd: parent })
```
eeeqeee, line 28:
```suggestion
    execSync(`git clone ${url} --depth 1 ${name}`, { cwd: parent })
```

libs/plugin/src/executors/sync-repo/executor.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const hashPath = path.format({ ...path.parse(cwd), base: '', ext: '.hash' })
+  fs.writeFileSync(hashPath, hash)
+  return { success: true }
+}
+
+export const getLocalRepoHash = (cwd: string): string =>
+  `${execSync(`git rev-parse -q HEAD`, { cwd })}`.trimEnd()
+export const getRemoteRepoHash = (url: string): string =>
+  `${execSync(`git ls-remote -q ${url} HEAD`)}`.replace(/\s+HEAD\s*$/, '')
```
eeeqeee, line 43:
```suggestion
  `${execSync(`git rev-parse HEAD`, { cwd })}`.trimEnd()
export const getRemoteRepoHash = (url: string): string =>
  `${execSync(`git ls-remote ${url} HEAD`)}`.replace(/\s+HEAD\s*$/, '')
```

libs/plugin/src/executors/sync-repo/executor.ts
lantua: The output texts are gonna be cached. So I'm not sure if we should include volatile texts from git. If you insist, I can un-quiet them.

nguyentvan7 (approved):

PR #1232, opened 2023-07-12
Add dmgNode for Lisa's Burst Cast DMG

PolomoPT:
couldn't help but to take a yoimiya dmgNode out of a conditional and to add a percent() to a candace one, sry.


apps/frontend/src/app/Data/Characters/Lisa/index.tsx
```diff
@@ -97,6 +105,10 @@ const dmgFormulas = {
     press: dmgNode('atk', dm.skill.press, 'skill'),
   },
   burst: {
+    summon: customDmgNode(
```
nguyentvan7, line 108: Shouldn't be able to aggravate, but nothing we can do now I think

nguyentvan7 (approved):

PR #1233, opened 2023-07-12
Fix PR Deployment Script

nguyentvan7:
Wip


eeeqeee (commented):
Can you pr this from a fork to test? afaik `permissions` can only restrict, not give more permissions

PR #1236, opened 2023-07-13
Move the remaining `pipeline` into `genfile`

lantua:
Move pipeline scripts for `dm-localization`, `localization`, and `silly-wisher-names` into `genfile` target and remove `pipeline` target altogether


lantua:
Merged into #1238 instead

PR #1238, opened 2023-07-14
Make tsconfig more consistent across projects

lantua:
Since all non-`tsc` compilation uses the source code directly, we should ensure that the `tsconfig` is uniform across different projects. The main change here is switching from the `CommonJS` to the `ESNext` module.


lantua:
We have to move the remaining genfile scripts into the executor since the old structure requires `CommonJS`

nguyentvan7 (approved):

PR #1243, opened 2023-07-16
Add conditional conditional support

### LINKED ISSUE #1049: Allow a conditional number of conditional states
For example, Yanfei or Ayato can have an increased number of stacks for their abilities if they are a higher constellation. We currently just show the absolute max number of stacks, even if the character does not meet that constellation requirement.

So we should be able to change the range of states shown to the user, based on character info, like conditional state or number of constellations.

If implemented now, this would be similar to how fields can display a number, or give a function that takes a UIData, performs some action on it, then outputs a number based on that

image

nguyentvan7:
* Resolve #1049 
* Yanfei, Mika and Ayato now both only show the max number of stacks possible based on their constellation value
chrome_Z92ZXgwUxy


frzyc (dismissed):
This looks to be working. Ideally, we wouldn't go through so much code churn to implement a requirement like this. This should be a design consideration for Pando Conditionals in the foreseeable future.

apps/frontend/src/app/Components/Conditional/ConditionalSelector.tsx
```diff
@@ -111,16 +119,23 @@ function ExclusiveConditionalSelector({
...
     <SqBadge color="secondary">Not Active</SqBadge>
   )
   const condName = getCondName(conditional.name)
 
+  // Conditional value is present, but there is no corresponding state.
+  // Can be caused by old db, or by changing constellation
+  // or changing a condition that disables another state
+  if (!state && conditionalValue) {
+    // Clear out the value
+    setConditional(undefined)
```
frzyc, line 136: This is "UI data validation", and something that I dislike happening, because it means that there is a point where the data is "invalid" in the database side, and you'd need to render the UI to correct it. In my opinion, I much prefer to wrap the conditional data in a "range" node, so that validation is actually built into the system itself. I understand that not to be possible in WR, so this is another design consideration for Pando Conditionals.

nguyentvan7: We can do both if you want. I didn't add the range node because it was janky in UI rendering. But with UI validation, that means we can fix the UI side and also not require the UI validation to fix a bad db, until the user loads the UI anyways

nguyentvan7:
Changed behavior.
Before: Making the state invalid would invalidate the conditional and set it to undefined
Now: Making the state invalid will only visually remove the conditional value. If you make the state valid again, it will still remember your previous choice.

Make Mika C6. Set detector stacks to 4. Change to C5. See that detector stacks is set to "Not Set". Change back to C6. See that detector stacks went back to 4.

apps/frontend/src/app/Components/Conditional/ConditionalSelector.tsx
nguyentvan7: Removed UI data validation and opted for validation on the sheet side, with visual fixes to the selector. Also fixed a bug where I was changing state from cross component, causing error on react

frzyc (approved):

frzyc (approved):

frzyc (approved):

PR #1253, opened 2023-07-22
Misc. Sheet Changes (some more important than others)

PolomoPT:
Add "duration" to many artifact sets' 4pc effects.
Change Lost Prayer's cond name (new sheet string)
Removes a redundant string from 4pc Traveling Doctor
Makes Vourukasha Glow's 4pc effect display even with 0 stacks


apps/frontend/src/app/Data/Artifacts/VourukashasGlow/index.tsx
```diff
@@ -59,10 +91,10 @@ const sheet: IArtifactSheet = {
             name: st('stack', { count: stack }),
             fields: [
               {
-                node: set4_skill_dmg_,
+                node: set4_stacks_skill_dmg_,
               },
               {
-                node: set4_burst_dmg_,
+                node: set4_stacks_burst_dmg_,
```
nguyentvan7, line 97: Should we add duration?

nguyentvan7 (commented):
Looks good to me other than the format CI failure

nguyentvan7 (approved):

PR #1276, opened 2023-08-28
Remove bad sort

nguyentvan7:
Fixes bad sorts like this:
image
Code was originally there for hydro traveler conditional, since the values were [0.5, 1, 1.5, etc], it was incorrectly sorted for some reason without that call. But we changed the conditional format to not use that, so this change was never needed, and is just a bug now.


frzyc (dismissed):
A bit of a nitpick for cornercases (Which might not exist)
Objects do not guarantee order. In most cases, they are by insert over, however, if the keys are "number-like" they are sorted ascendingly. 
Perhaps if we need to guarantee order, a `Map` should be used here. 
Could be a consideration for Pando Conditionals.

nguyentvan7:
Yea that exact cornercase was happening with hydro traveler conditionals with states 0.5, 1, 1.5, etc. Forgot what the order was, but it was not what was expected

frzyc (approved):

PR #1282, opened 2023-09-08
Update Scanner Tab

PolomoPT:
- Update to 4.0
- Remove Amenoma
- Mention new AdeptiScanner features


libs/gi-localization/assets/locales/en/page_scanner.json
```diff
@@ -25,10 +25,6 @@
     "p1": "Scans all artifacts in your inventory. Has a manual scanning mode.",
```
nguyentvan7, line 25: Can you add Adepti's new features with weapon scanning and enka-based character scanning?

PolomoPT: oh, right, that.

nguyentvan7 (approved):

PR #1285, opened 2023-09-24
Add more deployments

nguyentvan7:
* Adds automatic deployment for gi-frontend, sr-frontend and frontend apps, based on files edited
* Removed manual deployment option
* Fixed deployment so forked PRs can deploy properly
  * [x] We should ensure that forked PR actions do not run automatically, and require approval first, for safety
* Updated deployment comment to be a log, rather than a single line
* Fixed some various prod-build issues in gi-frontend that were surfaced by the new deployment action


nguyentvan7:
Here is an example. All 3 are deployed because git diff is being weird after a merge, but it was only deploying gi-frontend earlier when the changes were solely within the gi-formula module, as evidenced by the deployment comment log.

https://github.com/nguyentvan7/genshin-optimizer/pull/18

eeeqeee (changes requested):

eeeqeee (approved):

PR #1288, opened 2023-09-29
Add a percentage of difference to FieldDisplay for some compared values

ChemiKyle:
When looking at my results I usually convert to a percentage difference in my head, some of my multi-target configs are getting high enough that I'm dropping orders of magnitude; I figure I'm not alone in this and wanted to add it to the UI.

Examples of the new UI:

1. Optimizing artifacts:
image

2. Theorycrafting:
image

I tried to follow what appear to be your style preferences - I did exceed the 80 character line limit by 1 for `pctDiff`, that isn't in your `.editorconfig` so I imagined you might not care much.

Intended to exclude values that are already percentages, or were previously 0.


apps/frontend/src/app/Components/FieldDisplay.tsx
```diff
@@ -125,6 +126,11 @@ export function NodeFieldDisplay({
             {' '}
             {diff > 0 ? '+' : ''}
             {valueString(diff, node.info.unit, node.info.fixed)}
+            {(node.info.unit != "%" && oldValue != 0) ? (
+              " (" + pctDiff + ")"
```
nguyentvan7, line 130: Can you use template literals please?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

nguyentvan7:
Will have to wait for @frzyc to get back. We have been iffy on adding % diff because it can be misleading, but I guess at the same time it can be just as misleading as DMG numbers, and also just as helpful

apps/frontend/src/app/Components/FieldDisplay.tsx
ChemiKyle: Done in [582d259](https://github.com/frzyc/genshin-optimizer/pull/1288/commits/582d259fc51f12597ff75674a5c0fbfe5f7b83f6)

frzyc:
Discord thread for discussion https://discord.com/channels/785153694478893126/1157601858001567834

nguyentvan7 (approved):

PR #1297, opened 2023-10-08
migrate to nx 16.10.0

frzyc:
- Ran `nx migrate latest`
- Remove tree-shaking hack in `apps\frontend\webpack.config.js`
  - verified bundle size to be the same using [Webpack Bundle Analyzer](https://github.com/frzyc/genshin-optimizer/wiki/Some-code-snippets-to-remember#webpack-bundle-analyzer)


nguyentvan7 (approved):

nx.json
```diff
@@ -103,7 +103,8 @@
       "!{projectRoot}/.eslintrc.json",
       "!{projectRoot}/**/?(*.)+(spec|test).[jt]s?(x)?(.snap)",
       "!{projectRoot}/tsconfig.spec.json",
-      "!{projectRoot}/jest.config.[jt]s"
+      "!{projectRoot}/jest.config.[jt]s",
+      "!{projectRoot}/src/test-setup.[jt]s"
```
frzyc, line 107: @lantua this was generated from the migration script, not sure if this changes any testing setup we had existing

lantua: Should be fine. At worst, it'll add more files to bundled `dist`. Plus we don't have a file matching this pattern, so it should be benign.

PR #1313, opened 2023-10-23
UI lib refactoring

frzyc:
## Describe your changes

 - Some UI refactoring to make it easier to leverage some of the existing components in `frontend` to be used by other apps/libs
 - modularize `Theme`

## Testing/validation

Run `yarn run mini-ci`


libs/gi-ui/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "gi-ui",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "libs/gi-ui/src",
```
nguyentvan7, line 4: Please add this to ignore-paths in https://github.com/frzyc/genshin-optimizer/blob/master/.github/workflows/deploy-sro-pr.yml

nguyentvan7 (changes requested):

apps/gi-frontend/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "types": ["vite/client", "vitest"],
     "module": "ESNext",
     "resolveJsonModule": true,
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "exactOptionalPropertyTypes": false
```
nguyentvan7, line 10: Why?

tsconfig.base.json
```diff
@@ -35,12 +35,19 @@
...
       ],
       "@genshin-optimizer/gi-formula": ["libs/gi-formula/src/index.ts"],
+      "@genshin-optimizer/gi-frontend-gql": [
+        "libs/gi-frontend-gql/src/index.ts"
+      ],
       "@genshin-optimizer/gi-good": ["libs/gi-good/src/index.ts"],
       "@genshin-optimizer/gi-localization": [
         "libs/gi-localization/src/index.ts"
       ],
+      "@genshin-optimizer/gi-prisma-schema": [
```
nguyentvan7, line 45: what is this and gi-frontend-gql? They don't seem to be in this PR

libs/ui-common/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "jsx": "react-jsx",
     "esModuleInterop": false,
-    "jsxImportSource": "@emotion/react"
+    "jsxImportSource": "@emotion/react",
+    "exactOptionalPropertyTypes": false
```
nguyentvan7, line 6: why pt2

libs/gi-ui/src/App.scss
```diff
@@ -0,0 +1,16 @@
+/* For Rarity background*/
```
nguyentvan7, line 1: Should we delete the frontend/app.scss and move it to use this? or is that already done if we use the gi-ui components? I just don't want to have duplicated css where if we change it, we need to change it in 2 places

tsconfig.base.json
frzyc: oh weird, I was sure i removed these

apps/gi-frontend/tsconfig.json
frzyc: with `exactOptionalPropertyTypes` on true, a lot of properties needs to be super defined:
Like you cant do 
```
type Prop={
  something?:string
}
function Com({something}:Prop){...}

// will typeerror out saying string|undefined can't be used
<Com something={cond?"string":undefined} />
```

you will have to explicitly type it to be 
```
type Prop ={
  something?:string|undefined
}
```

libs/gi-ui/src/App.scss
frzyc: I don't think any of the frontend is using gi-ui component for this reason.
For frontend to use gi-ui, I think we'd need to amend app.scss, and also update `Theme` with the modular theme.

apps/gi-frontend/tsconfig.json
nguyentvan7: If you think that level of safety isn't needed then I guess it's fine to remove

frzyc: I'll see if i can refactor this option away. I do agree that this is technically more safety, but I remember i struggled a lot with this when i move the components over. (A lot of libs don't have this option enabled when they wrote their typings, which makes "legal" code have red squiggles)

frzyc: yeah, looks like mui does not support this option. I think any lib with mui components will need this. option

libs/gi-ui/src/App.scss
frzyc: ... or just change implementation to use the new `content-` palettes

nguyentvan7 (approved):

Issue #1264, opened 2023-08-09
Modules cannot be imported by `vite` and `webpack` simultaneously


lantua:
Currently, the `webpack` and `vite` modules directly import the source code. Therefore, whenever a module `A` of either type imports module `B`, the source code of `B` must conform to the module `A` configuration.

This poses two problems.
1. If module `A` configuration is more restricted than module `B`, then the compilation may fail, and
1. If module `B` queries an env variable, it needs to use the method compatible with module `A`.

For the first problem, we can ensure that all `libs` modules are more restricted than `apps` modules, which is currently the case (at the time of writing).

The second problem is trickier because the `webpack` uses `process.env['NODE_ENV']` to access runtime env, while `vite` uses `import.meta.env.*`. The two methods are incompatible, so any module imported by both `vite` and `webpack` modules cannot query the env variable. Conversely, any module intended to do so will be locked to either `vite` or `webpack`.

nguyentvan7:
https://discord.com/channels/785153694478893126/819643218446254100/1171934532899971172

We should add a config that consumer can set on the calc object that specifies logging level or if its in development mode, etc. This will allow us to be framework agnostic

PR #1332, opened 2023-11-10
Make `pando` debug-check explicit

lantua:
## Describe your changes

Instead of using `process.env[NODE_ENV]`, which is framework dependent (see #1264), this moves the debug check into an explicit variable, which can be set by downstream users using `setDebugMode`.

The original design is to have the `isDebug` flag in `calc`, but there are multiple free functions accessing this env variable as well, so I move it into module-wide variable.

Currently, it can only be toggled on or off, but it can be separate into "calculator debug" and "tag database debug" if needed.

## Issue or discord link

https://discord.com/channels/785153694478893126/819643218446254100/1171934532899971172

## Testing/validation

n/a

### DIFF
--- a/libs/pando/src/index.ts
+++ b/libs/pando/src/index.ts
@@ -1,2 +1,3 @@
 export * from './node'
 export * from './tag'
+export { setDebugMode } from './util'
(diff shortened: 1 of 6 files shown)


lantua:
Address, for `pando` specifically, #1264 (which may still be ongoing for other modules).

frzyc (approved):

PR #1347, opened 2023-11-19
Stream changes from go-next proj to master

frzyc:
## Describe your changes

Quite a lot of changes have went into the https://github.com/frzyc/genshin-optimizer/tree/frzyc/go-next branch in dependent libs. This change seeks to bring parity between the branch and master.


libs/gi-art-scanner/src/lib/enStringMap.tsx
```diff
@@ -0,0 +1,40 @@
+import type { ElementWithPhyKey } from '@genshin-optimizer/consts'
```
nguyentvan7, line 1: nit: should be .ts?

libs/gi-formula-ui/.babelrc
```diff
@@ -0,0 +1,13 @@
+{
```
nguyentvan7, line 1: What is the purpose of this library? As far as I understand, there won't be any apps that share this? Or is it just so the pathing for imports looks neater?

frzyc: its actually used by go-next since its using the pando UI. I assumed if we are still supporting gi-frontend it would use this as well.

nguyentvan7: isn't go-next replacing gi-frontend?

frzyc: yeah, basically the intention is for this lib to host UI stuff that uses gi-formula,

PR #1359, opened 2023-12-03
Character talent tab translations

liaojh1998:
## Describe your changes

* Resolves translation comments in #1358.
* Standardized translations to languages in Genshin by the following:
  * Changes "Normal/Charged Attack" to just "Normal Attack". All translations use "Normal Attack".
  * All "Constellation Lv." follows the same translations as Genshin.
  * All "Talent Lv." follows the same translations as Genshin.
  * Instead of "Unlocked at Ascension #", use the same translations in Genshin for "Unlocks at Character Ascension Phase #".
  * Instead of "Unlocked by Default", modified to "Passive Talent" and used the same Genshin translations for "Passive Talent".

## Testing/validation

image
image
image
image


liaojh1998:
Blocking until #1358 is submitted.

libs/gi-localization/assets/locales/en/page_character.json
```diff
@@ -57,5 +57,40 @@
   "tabTeambuff": {
     "team_reso": "Team Resonance",
     "resonance_tip": "The party must be full with 4 characters in order for Elemental Resonance to take effect."
+  },
+  "tabTalent": {
+    "constellation": "Constellation",
+    "constellation_lv": {
```
frzyc, line 63: You can use https://www.i18next.com/translation-function/interpolation to minimize the churn in `constellationLvl` and in `talentLvl`

frzyc (changes requested):
Hey @liaojh1998, I appreciate your work here on the translations. We are currently putting this site on maintenance mode, in favor of working on a new UI with a new calculation engine. so I am generally not scoping time for translations for this UI. However, I want to make sure that your time is respected, and I will work with you to put this translations in, especially since you've spent the time to update the translations.

liaojh1998:
Thanks, I'll hold off on any translation stuff in other changes. Sorry for the inconvenience here.

PolomoPT (changes requested):
I take it those translations were datamined, yeah? You should have them out of the localizations file, then.

Get the string ID and list it under libs/gi-dm-localization/src/executors/gen-locale/lib/Data/sheet.ts, and then reference that instead. We reference stuff in that file typically with `stg`, you can find it in most character pages.

TL;DR: gi-localization is for things that need manual translation, such as conditionals. these extracted strings should be in gi-dm-localization, including the English ones.

frzyc (approved):

nguyentvan7 (approved):
Deployed and tested with onetime deployment pipeline. Will clean after merge

PR #1386, opened 2023-12-26
Convert DMs to submodules

eeeqeee:
## Describe your changes

Convert the datamines to submodules to ensure consistent builds and resiliency to renaming of branches

## Issue or discord link

- https://discord.com/channels/785153694478893126/819643218446254100/1188992490943021156


package.json
```diff
@@ -7,7 +7,7 @@
     "frontend": "nx serve frontend",
     "build-all": "nx run-many -t build",
     "mini-ci": "nx format:write && nx affected -t test,lint --max-warnings=0",
-    "reload-dm": "nx run-many -t load-dm --skip-nx-cache"
+    "reload-dm": "git submodule update --init"
```
nguyentvan7, line 10: Seems like this might clone the whole DM repo, rather than just the head?

libs/plugin/src/executors/sync-repo/executor.ts
```diff
@@ -1,52 +0,0 @@
...
-    fs.mkdirSync(parent, { recursive: true })
-    execSync(`git clone ${url} --depth 1 ${name}`, { cwd: parent })
-  }
-
-  // Compute hash
-  const localHash = getLocalRepoHash(cwd)
-  console.log(`Synced ${name} with hash ${localHash}`)
-
-  const hashPath = path.format({ ...path.parse(cwd), base: '', ext: '.hash' })
-  fs.writeFileSync(hashPath, localHash)
```
frzyc, line 45: So I am not sure how submodules are updated, but basically its extremely expensive to store the entire dm repo in cache. This makes sure that we export the hash so that instead of checking for the entire repo, it only checks for a single hash file. 
in this case, how do we ensure that builds on the same submodule version is cached, and how do we ensure that builds are updated when submodule is updated?

eeeqeee: Theres conflicting documentation on whether NX natively supports submodules, so I added back the executor which should ensure it caches properly

lantua (approved):

PR #1406, opened 2024-01-02
Update SRO context usage/UI issues, add useLightCone hook

nguyentvan7:
## Describe your changes
* Update SRO context usage to be simpler
* Fix some minor UI issues
* Add useLightCone hook for later usage

## Issue or discord link

-

### DIFF
--- a/libs/sr-ui/src/Context/DatabaseContext.tsx
+++ b/libs/sr-ui/src/Context/DatabaseContext.tsx
@@ -1,14 +1,25 @@
 import { DBLocalStorage, SandboxStorage } from '@genshin-optimizer/database'
 import { SroDatabase } from '@genshin-optimizer/sr-db'
 import type { ReactNode } from 'react'
-import { createContext, useCallback, useMemo, useState } from 'react'
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useMemo,
+  useState,
+} from 'react'
 
 type DatabaseContextObj = {
   databases: SroDatabase[]
   setDatabase: (index: number, db: SroDatabase) => void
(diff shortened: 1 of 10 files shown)


libs/sr-ui/src/Context/DatabaseContext.tsx
```diff
@@ -1,15 +1,26 @@
...
 
 type DatabaseContextObj = {
   databases: SroDatabase[]
   setDatabase: (index: number, db: SroDatabase) => void
   database: SroDatabase
 }
+
 export const DatabaseContext = createContext({} as DatabaseContextObj)
 
+export function useDatabaseContext() {
```
frzyc, line 20: Is this better than 2 imports? I guess it doesnt really matter, but I think you should try to enforce one vs the other. If you want to use this hook exclusively, do not export `DatabaseContext` so only this hook can be imported and used.

libs/sr-ui/src/Hook/useLightCone.ts
```diff
@@ -0,0 +1,24 @@
...
+  useEffect(
+    () => setLightCone(database.lightCones.get(lightConeId)),
+    [database, lightConeId]
+  )
+  useEffect(
+    () =>
+      lightConeId
+        ? database.lightCones.follow(
+            lightConeId,
+            (_k, r, v) => r === 'update' && setLightCone(v)
```
frzyc, line 18: I wonder if you should also listen to the "deletion" event as well, since that would also need to update UI?

frzyc (approved):
Minor non-blocking comments

frzyc (approved):
:shipit:

PR #1420, opened 2024-01-06
Add optimization with workers to SRO

nguyentvan7:
## Describe your changes

* Add optimization target selector
* Add optimization process
* Add worker-in-worker spawning for optimization
* Refactor some UI code for build display

## Issue or discord link

- https://github.com/users/frzyc/projects/2/views/1?pane=issue&itemId=47251672
- https://github.com/users/frzyc/projects/2/views/1?pane=issue&itemId=47251687

## Testing/validation

- Tested optimizing, results looked correct from my eye.


libs/sr-opt/src/parentWorker.ts
```diff
@@ -0,0 +1,177 @@
...
+          switch (data.resultType) {
+            case 'initialized':
+              // Worker is initialized; start optimizing
+              worker.postMessage({ command: 'start' })
+              break
+            case 'results':
+              numBuildsComputed += data.builds.length
+              results = results.concat(data.builds)
+              // TODO: This is slow, might need to think of something else
+              if (results.length > maxResults) {
```
nguyentvan7, line 110: Should we keep a sorted array and have customer inserter/merger? Not sure if it would be any faster

frzyc: I think this is fine, and looks like this happens in worker space, and isnt a hot zone. I'd say keep this simple is better than writing something hard to debug/maintain.

libs/sr-opt/src/optimize.ts
```diff
@@ -0,0 +1,83 @@
...
+  relicsBySlot: Record<RelicSlotKey, ICachedRelic[]>,
+  numWorkers: number,
+  setProgress: (progress: ProgressResult) => void
+) {
+  // Spawn a parent worker to compile nodes, split/filter relics and spawn child workers for calculating results
+  const worker = new Worker(new URL('./parentWorker.ts', import.meta.url), {
+    type: 'module',
+  })
+
+  // Wait for parent worker to report complet
```
lantua, line 35: Typo

libs/sr-opt/src/optimize.ts
```diff
@@ -0,0 +1,83 @@
...
+  const buildResults = await new Promise<BuildResult[]>((res, rej) => {
+    worker.onmessage = ({ data }: MessageEvent<ParentMessage>) => {
+      switch (data.resultType) {
+        case 'progress':
+          setProgress({
+            numBuilds: data.numBuilds,
+            numBuildsCompute: data.numBuildsComputed,
+            waitingForResults: data.sendingResults,
```
lantua, line 43: Maybe we can match the variable names, e.g. `computing`, so we can just send `data` over?

libs/sr-opt/src/parentWorker.ts
lantua: We can set threshold to a value much larger than `maxResult`, and close to memory limit (and make sure it's still larger than `maxResult`). Otherwise, this will trigger on every loop after the first time, which is very slow.

libs/sr-opt/src/parentWorker.ts
```diff
@@ -0,0 +1,177 @@
...
+
+async function start({
+  relicsBySlot,
+  detachedNodes,
+  numWorkers,
+  maxResults,
+}: ParentCommandStart) {
+  // Step 3: Optimize nodes, as needed
+  const flatNodes = flatten(detachedNodes)
+  const combinedNodes = combineConst(flatNodes)
```
lantua, line 79: It might be better to keep reusing the name `detachedNodes`, i.e. `detachedNodes = combineConst(...)`. The result is supposed to be "the same computation but faster". If this comes from the example code, maybe change that too.

libs/sr-opt/src/parentWorker.ts
lantua: Also, if we keep track of the largest build after each sort, and use that to filter suboptimal builds, that should prevent `results` from growing too fast if good builds are found early.

lantua (dismissed):
Optimization logic lgtm

libs/sr-opt/src/childWorker.ts
```diff
@@ -0,0 +1,154 @@
...
+                  head: head.id,
+                  hand: hand.id,
+                  feet: feet.id,
+                  body: body.id,
+                  sphere: sphere.id,
+                  rope: rope.id,
+                },
+              })
+              // Only sort and slice occasionally
+              if (builds.length > MAX_BUILDS_TO_SEND) {
```
frzyc, line 120: nit: can hoist as a function under line 91, since its duped below.

frzyc (dismissed):

PR #1421, opened 2024-01-06
fix useEquippedRelics

frzyc:
## Describe your changes

Fix some general issues with the `useEequippedRelics` hook

## Issue or discord link

- https://cdn.discordapp.com/attachments/1183809278209970217/1193073948963901460/image.png?ex=65ab63b3&is=6598eeb3&hm=9a29ace0ebec2c0fd8b7923d10d5285632fb7cbc3601fcafb21a6734b263052d&

### DIFF
--- a/apps/sr-frontend/src/app/Character.tsx
+++ b/apps/sr-frontend/src/app/Character.tsx
@@ -27,7 +27,7 @@ export default function Character() {
   const { characterKey } = useCharacterContext()
   const character = useCharacter(characterKey)
   const charReducer = useCharacterReducer(characterKey)
-  const relics = useEquippedRelics(characterKey)
+  const relics = useEquippedRelics(character?.equippedRelics)
 
   const { calc } = useCalcContext()
   const member0 = convert(selfTag, { member: 'member0', et: 'self' })
@@ -67,12 +67,13 @@ export default function Character() {
                   {Object.values(relics).map((relic, index) => {
                     return (
                       <Grid item key={index} xs={1}>
-                        relic ? (
-                        <RelicCard relic={relic} />) : (
-                        <Box>
-                          <Typography>Empty</Typography>
-                        </Box>
-                        )
+                        {relic ? (
+                          <RelicCard relic={relic} />
+                        ) : (
+                          <Box>
+                            <Typography>Empty</Typography>
+                          </Box>
+                        )}
                       </Grid>
                     )
                   })}
--- a/libs/sr-ui/src/Hook/useEquippedRelics.ts
+++ b/libs/sr-ui/src/Hook/useEquippedRelics.ts
@@ -26,14 +26,15 @@ export function useEquippedRelics(
   useEffect(() => {
     setRelics(objMap(relicSlotIds, (id) => database.relics.get(id)))
     const unfollows = Object.values(relicSlotIds).map((relicId) =>
-      database.relics.follow(
-        relicId,
-        (_k, r, v: ICachedRelic) =>
-          (r === 'update' || r === 'remove') &&
-          setRelics((relics) => ({ ...relics, ...{ [v.slotKey]: v } }))
-      )
+      database.relics.follow(relicId, (_k, r, v: ICachedRelic) => {
+        if (r === 'update')
+          setRelics((relics) => ({ ...relics, [v.slotKey]: v }))
+        // remove event returns the deleted obj
+        if (r === 'remove')
+          setRelics((relics) => ({ ...relics, [v.slotKey]: undefined }))
+      })
     )
-    return unfollows.forEach((unfollow) => unfollow())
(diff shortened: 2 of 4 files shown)


apps/sr-frontend/src/app/Character.tsx
```diff
@@ -67,12 +67,13 @@ export default function Character() {
                   {Object.values(relics).map((relic, index) => {
                     return (
                       <Grid item key={index} xs={1}>
-                        relic ? (
```
frzyc, line 70: The issue there is that you are printing `relic ? (` instead of using it as js logic. need to wrap `{}`

libs/sr-ui/src/Hook/useEquippedRelics.ts
```diff
@@ -1,54 +1,40 @@
...
+import type {
+  ICachedRelic,
+  ICachedSroCharacter,
+} from '@genshin-optimizer/sr-db'
+import { objKeyMap, objMap } from '@genshin-optimizer/util'
+import { useEffect, useMemo, useState } from 'react'
 import { useDatabaseContext } from '../Context'
-import { useCharacter } from './useCharacter'
 
-export function useEquippedRelics(characterKey: CharacterKey | '') {
```
frzyc, line 7: Since most of the time you are getting `character` outside of this hook already in the first place, you are actually storing `character` twice. Since this hook is intended to handle `equippedRelics`,concern is shifted upwards.

libs/sr-ui/src/Hook/useEquippedRelics.ts
```diff
@@ -1,54 +1,40 @@
...
+export function useEquippedRelics(
+  equippedRelics: ICachedSroCharacter['equippedRelics'] | undefined
+) {
+  const relicSlotIds = useMemo(
```
frzyc, line 16: a memoized obj to ensure the object is always defined

libs/sr-ui/src/Hook/useEquippedRelics.ts
```diff
@@ -1,54 +1,40 @@
...
-      setRelics(
-        Object.fromEntries(
-          Object.entries(character?.equippedRelics).map(
-            ([slotKey, relicId]) => [slotKey, database.relics.get(relicId)]
-          )
-        ) as Record<RelicSlotKey, ICachedRelic | undefined>
-      )
-    )
-  }, [database, characterKey, character?.equippedRelics])
+  >(() => objMap(relicSlotIds, (id) => database.relics.get(id)))
```
frzyc, line 24: you can initialize `relics` with the actual relics on initialization, so you wouldnt need another update cycle with `useEffect` to populate it.

libs/sr-ui/src/Hook/useEquippedRelics.ts
```diff
@@ -1,54 +1,40 @@
...
-        (Object.fromEntries(
-          Object.entries(character?.equippedRelics).map(
-            ([slotKey, relicId]) => [
-              slotKey,
-              { ...database.relics.get(relicId) },
-            ]
-          )
-        ) as Record<RelicSlotKey, ICachedRelic | undefined>)
-    )
-    return (
```
frzyc, line 25: The `return` value of useEffect is called on next iteration of useEffect. I think your intention here is to run it in the "current" iteration

libs/sr-ui/src/Hook/useEquippedRelics.ts
```diff
@@ -1,54 +1,40 @@
...
-  }, [database, characterKey, character?.equippedRelics])
+  >(() => objMap(relicSlotIds, (id) => database.relics.get(id)))
 
   useEffect(() => {
-    const relicIds = character?.equippedRelics
-      ? Object.values(character.equippedRelics)
-      : []
-    if (relicIds) {
-      relicIds.forEach((relicId) =>
-        database.relics.follow(
```
frzyc, line 43: you are registering for the event listener with `follow`, but it actually returns a function to unfollow, which you are not using.

libs/sr-ui/src/Hook/useEquippedRelics.ts
```diff
@@ -1,54 +1,40 @@
...
+      database.relics.follow(
+        relicId,
+        (_k, r, v: ICachedRelic) =>
+          (r === 'update' || r === 'remove') &&
+          setRelics((relics) => ({ ...relics, ...{ [v.slotKey]: v } }))
       )
-    }
-  }, [character?.equippedRelics, characterKey, database, relics])
+    )
+    return unfollows.forEach((unfollow) => unfollow())
```
frzyc, line 36: proper use of `return` in `useEffect`. Will unfollow if equippedRelics are changed, which likely means a new character, who has a different subset of relics, and you need to unfollow updates from the previous subset of relics.

Issue #829, opened 2022-12-16
Add 4th "hit mode" option within multi-target to "follow global"


StainAE86:
Use case: There are times you want to optimize for crit hit on a multi-target but currently you have to make a copy of the target and swap every hit within it to crit then select this new target.

Allowing the user to select a 4th hitmode option (and possibly make it default?) that will just follow the global / main setting would be nice:

![image](https://user-images.githubusercontent.com/101696552/208178779-ecb9d7a9-cef2-4957-a989-fab11d51c985.png)

frzyc:
Could implement for hitmode toggle, but not reaction/infusion/conditional toggle

nguyentvan7:
Completed in #1433

PR #1433, opened 2024-01-13
"Follow Global" option for multi-target.

SirDerpyHerp:
## Describe your changes
- Added a "Follow Global" option for multi-target.
## Issue or discord link
- #829

## Testing/validation
image
image
image

### DIFF
--- a/apps/frontend/src/app/PageCharacter/CustomMultiTarget.tsx
+++ b/apps/frontend/src/app/PageCharacter/CustomMultiTarget.tsx
@@ -110,10 +110,7 @@ function validateCustomTarget(ct: unknown): CustomTarget | undefined {
   if (
     !hitMode ||
     typeof hitMode !== 'string' ||
-    (
-      !allHitModeKeys.includes(hitMode as HitModeKey) &&
-      hitMode !== 'global'
-    )
+    (!allHitModeKeys.includes(hitMode as HitModeKey) && hitMode !== 'global')
   )
     hitMode = 'avgHit'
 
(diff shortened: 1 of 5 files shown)


apps/frontend/src/app/Types/character.d.ts
```diff
@@ -21,7 +21,7 @@ import type { IConditionalValues } from './sheet'
 export interface CustomTarget {
   weight: number
   path: string[]
-  hitMode: HitModeKey
```
nguyentvan7, line 24: I'd prefer to have a `MultiOptHitModeKey` with 'global' added so it can simplify this, the validation code for multi-opts, and the UI code as well

```typescript
const allMultiOptHitModeKeys = [...allHitModeKeys, 'global']
```

apps/frontend/src/app/PageCharacter/CustomMultiTarget.tsx
```diff
@@ -110,7 +110,7 @@ function validateCustomTarget(ct: unknown): CustomTarget | undefined {
   if (
```
nguyentvan7, line 110: Should we adjust the default (line 91) to be 'global'?

nguyentvan7 (changes requested):

apps/frontend/src/app/Types/character.d.ts
SirDerpyHerp: I've implemented this now

apps/frontend/src/app/PageCharacter/CustomMultiTarget.tsx
frzyc: I'd say we make it opt-in for now, we can always make it the default later, if there is an overwhelming request for it to be

frzyc (approved):

nguyentvan7 (approved):

PR #1434, opened 2024-01-13
Fix SR data rounding

lantua:
## Describe your changes

SR numbers seem to use some floating-point type with precision between single-precision and double-precision, so we need to convert them to double-precision to match JS `number` type, similar to those in GO.

Simply calling `Math.fround` will fail on some numbers because of double rounding, from original number -> SR datamine and from SR datamine -> final value. Instead, the numbers from datamine are passed to `extrapolateFloat` as is, and are treated as single-precision values. This requires some updates to `extrapolateFloat`, which is also included in this PR.

## Issue or discord link

https://discord.com/channels/785153694478893126/1183809278209970217/1192203678334079146

## Testing/validation

- `gi-stats` remains unchanged with the new pipeline code
- `sr-stats` uses fewer digits for most stats

- [X] I have commented my code, in hard-to understand areas.
- [X] I have made corresponding changes to README or wiki.
- [X] For front-end changes, I have updated the corresponding English translations.
- [X] Ran `yarn run mini-ci` locally to validate format + lint.
- [X] If there were format issues, I ran `nx format write` to resolve them automatically.
- [X] If I have added a new library or app, I have updated the deployment scripts to ignore changes as needed


nguyentvan7 (approved):
Only looked at the datamine diffs, looks good to me

Issue #963, opened 2023-02-10
"Swap artifact" pop-up only shows 12 first artifacts with no indication that this isn't the whole list.


CaspianRoach:
The pop-up window for "Swap artifact" should either have a page control element or it should include a notice that it displays only first 12 artifacts that are valid for current filters, if this behaviour is intentional.

At the moment it is very easy to misunderstand and think that you don't have any more artifacts than the current cap of 12.

Pictured: bottom of the "Swap artifact" window, with no page control or limit reminder.

![image](https://user-images.githubusercontent.com/7204157/218175606-86285d3b-7521-44f1-8227-c831caa02c92.png)

CaspianRoach:
I know that it shows 12/408 and the sort on top of the filters in the top left part of the window, but this is ambiguous as this could also mean "only 12 artifacts are valid for currently selected filters".

Also, while selecting an artifact set in that window, the first part of the artifact count only refers to the currently 12 visible artifacts. Before clicking: 
![image](https://user-images.githubusercontent.com/7204157/218177075-595d95ed-d778-4487-87a8-2bfa443feaab.png)

it says 3 artifacts

After clicking: 
![image](https://user-images.githubusercontent.com/7204157/218177159-193d8248-5036-4ca5-b69e-9cea8a11c811.png)

it says 12, despite the fact that I have 28. (Checked it on the "Artifacts" page, which all works correctly. This miscalculation only happens on the Character -> Overview -> Swap Artifact)

StainAE86:
This is three issues then.

1) Add pagination to the swap artifact popup:
![image](https://user-images.githubusercontent.com/101696552/218178567-4e137a82-10a2-4fe6-950e-c10e70e042fc.png)

2) Add indication that the number is "showing" rather than "filtered":
![image](https://user-images.githubusercontent.com/101696552/218178844-0924cd43-305d-43df-a731-46c83b440e89.png)

3) Change the "total" number for the set/mainstat/etc filters to be per slot, rather than your total artifact inventory

CaspianRoach:
I think all 3 stem from lack of pagination and I'm assuming will automatically fix themselves when page controls will be added.

CaspianRoach:
Except for your suggestion 3, which I didn't even think about, you're right, with only Flowers selected, the dropdown should probably not include options like "Anemo DMG%"

StainAE86:
[...]

PR #1436, opened 2024-01-18
Add pagination to Swap Artifact UI and changed artifact totals to reflect only artifacts for slot being swapped

arquelion:
## Describe your changes
- Added pagination to the Swap Artifact UI so all artifacts for the slot being swapped are displayed across multiple pages instead of only the first 12
- Added "Showing x out of y/z Artifacts" text clarifying how many artifacts are being shown (x), how many are being filtered (y), and how many match the selected slot (z)
- Modified `ArtifactFilterDisplay` to check whether the slot filter is disabled (as it is when swapping artifacts) and to only increment totals for artifacts matching the specified slot if it is - This results in the artifact totals in the Swap Artifact UI accounting for only that slot instead of using the count for the entire inventory (e.g. you'll see 0 instead of 0/x for filter options that don't actually exist in a slot).

## Issue or discord link

- https://github.com/frzyc/genshin-optimizer/issues/963

## Testing/validation

image
image
image
image
image
image


arquelion:
A sidenote: I'm inclined to agree that pagination for artifact-related things should probably be its own component as I honestly lifted a lot of the pagination stuff here directly out of the code for the Artifacts page (and there are quite a number of code similarities between the two). I could try to do it as part of this PR, but I personally thought it better to submit this as-is to address #963 and leave the refactor as a separate item.

frzyc:
Looks like you need to run `nx format write` again

frzyc:
my bad, `nx format` is comparing to the `master` of the repo, since you are working the master of your fork, it is itself. In this case, the command is `nx format write --base=upstream/master` to compare with the main repo. I would recommend in the future, to create a feature branch on your fork.

arquelion:
Yeah I realized right before I was about to open the PR that I'd derped and forgotten to create a feature branch haha. I did wonder if that might've had something to do with it, so good to know.

nguyentvan7:
https://genshin-optimizer-prs.github.io/pr/1436/frontend/

frzyc (approved):

PR #1456, opened 2024-01-25
Pagination Refactor

arquelion:
## Describe your changes

- Add new component `PageAndSortOptionSelect` (in `Components/PageAndSortOptionSelect.tsx`) which is responsible for rendering the page number selector, the "Showing x out of y/z" text, and the Sort By button (which can be conditionally disabled through the `displaySort` boolean) - This new component takes three separate objects containing the props for `Pagination`, `ShowingItem`, and `SortByButton` as well as a boolean indicating whether to render the sorting options.
- Replaced all relevant sections in `PageArtifact`, `PageWeapon`, `PageCharacter`, and `ArtifactSwapModal` to use the new component instead

## Issue or discord link

- https://discord.com/channels/785153694478893126/819643218446254100/1197731809170051205

## Testing/validation

- I've manually checked the artifacts, weapons, and characters pages as well as the artifact swap UI to ensure page and sort option selection behavior is unchanged.


arquelion:
This is still WIP, but opening it as a draft PR now to get some feedback on a few things in particular:

- This implementation technically works, but I'm not entirely pleased with it as I'm passing a ton of props around due to `Pagination`, `ShowingItem`, and `SortByButton` all requiring a different set of props, not to mention the necessity of the `displaySort` prop in order to control whether the sorting options are shown at all. I tried handling this with the `children` prop but ran into issues with the artifact swap UI and the UI at the bottom of the artifact/weapon/character pages only having the page selector and the "Showing...." text and thus not passing anything in. I'd also considered the idea of determining as many of these props inside `PaginatedDisplay` as possible, but there are other variables in `PageArtifact` etc. that also require them for other calculations, and state didn't seem like the right solution. Passing the props as a few objects containing them makes the code a little nicer on the `PaginatedDisplay` end but still a nightmare to look at everywhere else. Being new to JS/React, is there an approach or JS feature I've missed that would make this much nicer to look at?
- `Sort By` button parity - We currently have the sorting options available at the top of the page but not the bottom, which means I currently pass in a bool to prevent the sort button from rendering at all (which consequently means a lot of default values since there's no guarantee any of the `SortByButton` props will be needed). Keep as-is or should I just render the button both at the top and the bottom of the page?

Commit history seems to have picked up the commits from the previous PR, I might've forgotten to rebase or something. Will resolve at some point before I mark this PR ready.

nguyentvan7:
You can somewhat simplify large prop passing by passing objects rather than individual fields. GeneralAutoComplete is an example of this, where the caller can pass a textFieldProps that will propagate down. It doesn't solve the caller having to pass this many options but it's slightly cleaner

I mention the above without actually reading what the props are yet, so it's possible the props could still be simplified in a different way

arquelion:
Props-wise it's number of pages, current page index, and `setPage()` for `Pagination`; a list of IDs for the arti/char/weaps to be displayed, the `Showing x out y` text, and the result from calling `useTranslation()` for `ShowingItem`; and the sort keys, currently selected sort option, whether to sort in ascending or descending order, and the two handlers for the two sets of sorting options for `SortByButton` plus a bool for whether or not to render the sorting stuff at all.

I tried to compute a lot of these values inside the new component since it didn't seem like most of these values were being used elsewhere, but the component which actually displays the artifact/weap/char cards requires the list of IDs being shown and it didn't seem like I could just return the list here, and computing that list requires the pagination-related stuff as well as some of the values needed for the "Showing" text anyway. I could abstract away at least the stuff for `SortByButton` if I hardcode things to only handle artifacts/weaps/chars and then pass in a prop that indicates which of the three it is (e.g. if "artifact" gets passed in, then it'll pass the artifact-relevant sort keys, etc. to the button component). Given artifacts/chars/weaps are the only things ever being sorted, maybe it'll be fine?

arquelion:
Update: I've split the huge list of individual props into three separate objects (each containing the necessary props for `Pagination`, `ShowingItem`, and `SortByButton`) along with the boolean controlling whether sorting options are displayed. It's still a lot of values being passed around, but at least the call sites look cleaner now. Still open to better implementation ideas though.

Marking this ready for review, though iirc #1457 separately added a pagination component to the artifact exclusion UI so I am fine waiting for that to be merged first then updating the pagination here.

arquelion:
Since it's been about a week since I marked this ready for review and PR 1457 has not been merged yet, I'd like to get this reviewed/landed and move on to adding pagination to the weapon swap UI in the interim (and deal with updating pagination in the artifact include/exclude UI later). @nguyentvan7 @frzyc I know you're both busy this week, but would you mind taking a look if you have a chance?

apps/frontend/src/app/PageArtifact/index.tsx
```diff
@@ -154,7 +146,7 @@ export default function PageArtifact() {
 
   const { artifactIdsToShow, numPages, currentPageIndex } = useMemo(() => {
     const numPages = Math.ceil(artifactIds.length / maxNumArtifactsToDisplay)
-    const currentPageIndex = clamp(pageIdex, 0, numPages - 1)
```
frzyc, line 157: lol its misspeeled - thanks for correcting it

apps/frontend/src/app/Components/PageAndSortOptionSelect.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onChangeAsc: (value: boolean) => void
+}
+
+export default function PageAndSortOptionSelect({
+  paginationProps,
+  showingTextProps,
+  displaySort = false,
+  sortByButtonProps = undefined,
+}: {
+  paginationProps: PaginationProps
```
frzyc, line 32: This is an interesting way to organize the props. I usually opt to have flatter props, but I like how you group these according to use. 
I think in this case, there are quite a lot of props, so this structure is more beneficial.

frzyc (approved):
Thanks a lot to your contribution. I've tested your code locally, and I can find no flaws 👍

frzyc:
Also, Feel free to ping us in discord(#go-development) next time you have a PR ready for review.

arquelion:
No worries, I did say I was fine waiting for the bulk artifact include/exclude to be merged first so I could bring that in and update the pagination, so I presumed that was why it hadn't been reviewed yet.

I did have a flat props list initially and picked that idea up from looking elsewhere after going off Van's comment up there, so credit to him haha

PR #1466, opened 2024-01-28
Optimize tc-opt

eeeqeee:
## Describe your changes

Only check feasiblity when `result > max`

## Issue or discord link

-

### DIFF
--- a/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabTheorycraft/optimizeTc.ts
+++ b/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabTheorycraft/optimizeTc.ts
@@ -212,23 +212,6 @@ export function optimizeTcUsingNodes(
         })
       if (x !== 'other') buffer[x] = substatValue(x, toAssign)
       bufferRolls[x] = toAssign
-      if (!alreadyFeasible) {
-        //check for distributed feasibility
-        const allRolls = allSubstatKeys.map((k) => [
-          k,
-          (existingRolls[k] ?? 0) + (bufferRolls[k] ?? 0),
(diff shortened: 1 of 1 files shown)


apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabTheorycraft/optimizeTc.ts
```diff
@@ -237,6 +220,23 @@ export function optimizeTcUsingNodes(
       }
       const result = results[0]
```
frzyc, line 221: why don't we change the flow a bit more? 

```
compute()
if(result <= max) return
check constraints return
check feasibility return
```

I guess my concern is the case where `compute` is more expensive than the feasibility check.

eeeqeee: I tested both Hu Tao `8 N2C` and Cyno `4TF Quickbloom` and they were both faster in this PR. On which multitarget do you think `compute` is more expensive?

frzyc: Can you try the same with moving the constraint check after the max check? wonder if that is faster as well

frzyc (approved):

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabTheorycraft/optimizeTc.ts
frzyc: To answer your previous question, I think we probably shouldn't cater to crazy mtargets, this solution is good.

PR #1496, opened 2024-02-13
Add option to directly delete artifacts from build results

### LINKED ISSUE #849: Allow deleting artrifacts directly from the build results
Use case: An artifact gets suggested that you've since foddered in-game. This way you can delete it from the database from the Optimize page and regenerate rather than making note of what it is and filtering/finding it on the artifact page then returning to the character's Optimize page.

arquelion:
## Describe your changes
- Added `deleteArtifact` callback to `CompareArtifactModal`
- Modified `ArtifactCard` call sites in `CompareArtifactModal` to pass `deleteArtifact` to `onDelete`
- Renamed `ExcludeButton` to `ArtInclusionButton` as this button is now a checkbox indicating whether to include the specified artifact in optimization and moved it onto its own line to avoid squishing the character selection box any further with the addition of the delete button
- Added new translation keys `includeArtifactTip` and `includeArtifactButton` and removed `excludeArtifactTip`, `exclude`, and `include` under `excludeArt` in `page_character_optimize.json`

## Issue or discord link

- Resolves #849 

## Testing/validation

`CompareArtifactModal` with these changes (as well as the Exclude Artifacts button showing the artifact is excluded)
image
Modal after deleting an artifact as well as the deletion being reflected in the build itself
image
Confirming that the artifact has indeed been removed from the database
image

### DIFF
--- a/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
+++ b/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
@@ -341,12 +341,12 @@ function ExcludeButton({ id }: { id: string }) {
     >
       <Button
         onClick={toggle}
-        color={excluded ? 'error' : 'success'}
+        color={excluded ? 'secondary' : 'success'}
         size="small"
-        startIcon={excluded ? <BlockIcon /> : <ShowChartIcon />}
+        startIcon={excluded ? <CheckBoxOutlineBlankIcon /> : <CheckBoxIcon />}
       >
         <span>
-          {t(excluded ? 'excludeArt.excluded' : 'excludeArt.included')}
+          {t`excludeArt.included`}
         </span>
       </Button>
     </BootstrapTooltip>
--- a/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
+++ b/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
@@ -31,7 +31,6 @@ import BootstrapTooltip from '../../../../../Components/BootstrapTooltip'
 import CardDark from '../../../../../Components/Card/CardDark'
 import CardLight from '../../../../../Components/Card/CardLight'
 import StatDisplayComponent from '../../../../../Components/Character/StatDisplayComponent'
-import ColorText from '../../../../../Components/ColoredText'
 import ModalWrapper from '../../../../../Components/ModalWrapper'
 import SqBadge from '../../../../../Components/SqBadge'
 import WeaponCardNano from '../../../../../Components/Weapon/WeaponCardNano'
@@ -295,7 +294,7 @@ function CompareArtifactModal({
               canEquip
               editorProps={{ disableSet: true, disableSlot: true }}
             />
-            <ExcludeButton id={newId} />
+            {newLoc && <ExcludeButton id={newId} />}
             {newLoc &&
               newLoc !== charKeyToLocCharKey(characterKey) &&
               allowLocationsState !== 'all' && (
@@ -335,11 +334,6 @@ function ExcludeButton({ id }: { id: string }) {
       title={
         <Box>
           <Typography>{t`excludeArt.excludeArtifactTip`}</Typography>
-          <Typography>
-            <ColorText color={excluded ? 'error' : 'success'}>
-              {t(excluded ? 'excludeArt.excluded' : 'excludeArt.included')}
-            </ColorText>
-          </Typography>
         </Box>
       }
       placement="top"
@@ -351,6 +345,9 @@ function ExcludeButton({ id }: { id: string }) {
         size="small"
       >
         {excluded ? <BlockIcon /> : <ShowChartIcon />}
+        <span>
+          {t(excluded ? 'excludeArt.excluded' : 'excludeArt.included')}
+        </span>
       </Button>
     </BootstrapTooltip>
   )
--- a/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
+++ b/apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
@@ -294,7 +294,7 @@ function CompareArtifactModal({
               canEquip
               editorProps={{ disableSet: true, disableSlot: true }}
             />
-            {newLoc && <ExcludeButton id={newId} />}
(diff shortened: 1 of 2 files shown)


apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -284,11 +289,12 @@ function CompareArtifactModal({
           <Box minWidth={320} display="flex" flexDirection="column" gap={1}>
             <ArtifactCard
               artifactId={newId}
+              onDelete={deleteArtifact}
               mainStatAssumptionLevel={mainStatAssumptionLevel}
               canEquip
               editorProps={{ disableSet: true, disableSlot: true }}
-              extraButtons={<ExcludeButton id={newId} />}
             />
+            {newLoc && <ExcludeButton id={newId} />}
```
frzyc, line 297: Should exclude button be conditional on `newLoc`?

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -344,6 +345,9 @@ function ExcludeButton({ id }: { id: string }) {
         size="small"
       >
         {excluded ? <BlockIcon /> : <ShowChartIcon />}
```
frzyc, line 347: move this icon to `starIcon` https://mui.com/material-ui/api/button/#button-prop-startIcon

frzyc (changes requested):
Minor question/request.

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
arquelion: Yes because otherwise it'll continue to be rendered even after the artifact is deleted, which doesn't match the behavior of the character/set exclusion buttons (and it'd be a little weird to have an exclude button for a non-existent artifact)

arquelion: Changed `newLoc` to `newArtifact`, done

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
arquelion: done

nguyentvan7 (commented):
To follow the pattern of our other buttons, it should show a checkbox, and the text should not change when the state is toggled. e.g. the text should be 
"Exclude artifact from optimization"
with an empty/filled checkbox as needed

arquelion:
There are a couple ways I could go about that, either make the box turn red indicating the artifact is currently excluded or make it turn green indicating the artifact is currently included (I'm leaning towards this one since it would match the style of the other two exclude buttons).

What should I do about translations though? "Exclude/Include this artifact from optimization" doesn't currently exist in the translation entries. Or should I make do with the excluded/included text for now and let the tooltip clarify the rest?

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -328,22 +332,18 @@ function ExcludeButton({ id }: { id: string }) {
...
     >
       <Button
         onClick={toggle}
-        color={excluded ? 'error' : 'success'}
+        color={excluded ? 'secondary' : 'success'}
         size="small"
+        startIcon={excluded ? <CheckBoxOutlineBlankIcon /> : <CheckBoxIcon />}
       >
-        {excluded ? <BlockIcon /> : <ShowChartIcon />}
+        <span>{t`excludeArt.included`}</span>
```
frzyc, line 346:
```suggestion
        <span>{excluded ?t`excludeArt.included` : t`excludeArt.included`}</span>
```

frzyc (changes requested):

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
arquelion: Why the ternary if it'll end up displaying the same text regardless of the value of `excluded`?

frzyc: Opps, copying error on my end. I was refering to https://github.com/frzyc/genshin-optimizer/blob/master/libs/gi/localization/assets/locales/en/page_character_optimize.json#L37
which is 
```
t`excludeArt.excluded`
```

arquelion: I did consider using `excludeArt.excluded`, but using `excludeArt.included` instead as the text means the button's behavior better matches the behavior of the character and set effect exclusion buttons (green if included, gray if excluded).

Though I could use the excluded text and invert the colors so green is excluded and gray is included. It seems a little weird in my head, but seeing as that's how the Exclude Artifacts button works anyway, it's consistent on that end.

nguyentvan7, line 346: prob dont need a span anymore

also the wording should be more like "Include this artifact in optimization"

nguyentvan7:
https://github.com/frzyc/genshin-optimizer/blob/c4356cedb2e32807eed121954fa9534b1a8f0df9/libs/gi/localization/assets/locales/en/page_character_optimize.json#L36

you can add a new key here for 'includeArtifactTip' and a new key for the button text too. if the old keys aren't used anywhere else, we should delete them as well

apps/frontend/src/app/PageCharacter/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
arquelion: done

arquelion:
I've made the necessary translation changes in `page_character_optimize.json` for the new tooltip and button text and removed the old ones, don't think anything else uses them (but I'll check this again to be extra sure). Testing/validation images have been updated again.

nguyentvan7 (approved):
strings/layout lgtm

frzyc (approved):

frzyc:
Thanks for your work!

PR #1512, opened 2024-02-18
Sort imports

### LINKED ISSUE #1510: Add a prettier plugin or linter step to organize imports automatically
This would make it independent of IDE and also able to be detected and thrown as an error in our CI

eeeqeee:
## Issue or discord link

- Fix #1510


nguyentvan7:
Is there a way to configure this with vscode to be on-save? I like to run Organize Imports on save, because it will also convert imports to be type-only when needed, and also removes extraneous imports. If there was somehow some plugin that could just replicate that, that would also be nice, but sounds a bit more complex

LuMiSxh:
> Is there a way to configure this with vscode to be on-save? I like to run Organize Imports on save, because it will also convert imports to be type-only when needed, and also removes extraneous imports. If there was somehow some plugin that could just replicate that, that would also be nice, but sounds a bit more complex

Adding the following lines to the projects vs-code `settings.json` should run aslant on save:
```json
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  }
```
(optionally add `"eslint.alwaysShowStatus": true` at the end as well)

PR #1574, opened 2024-02-28
Hotfix update missing migration data

HassnHamada:
## Describe your changes

Add `lastEdit` to the database with default value of `0` since it didn't exist before.

This was causing the build to fail

image

## Issue or discord link

- 


libs/gi/db/src/Database/migrate.ts
```diff
@@ -208,6 +208,7 @@ export function migrateGOOD(good: IGOOD & IGO): IGOOD & IGO {
         description: `Generated team due to database migration for GO version 10`,
         enemyOverride,
         teamCharIds,
+        ...{ lastEdit: 0 },
```
frzyc, line 211:
```suggestion
        lastEdit: 0,
```
Why does this need to be encapsulated?

frzyc (changes requested):

libs/gi/db/src/Database/migrate.ts
HassnHamada: MB. It doesn't need to.

PR #1576, opened 2024-02-29
Change PR deployment to build artifact + deploy

### LINKED ISSUE #1575: Convert PR deployment pipeline to be a build -> upload drop -> deploy pipeline
Currently our forked PRs cannot deploy due to lack of access to secrets in `pull_request` events. We can use `pull_request_target` events, but we cannot reliably get a merge SHA without hacky workarounds (#1327).

As a result, we are not seeing if forked PRs are able to actually build the site using our prod-rules.

Due to this, we should convert our deployment pipeline to just be a build pipeline. It will build the site, then zip the `dist/apps/frontend` folder along with some metadata. Then it will upload that as an artifact (free for public repos).

This step can be done for forked and local branches alike, and would verify the build step works on every PR.

After that, we can have an additional `workflow_run` call that will download the artifact, then deploy that using the metadata contained inside.
`workflow_run` calls from within a `pull_request` target will have access to all repo secrets, and so the deployment will actually be able to succeed for even forked repos. Inside this, we can also have the comment/deployment log maintenance, as that requires secrets.

tl;dr: move to build -> upload then download -> deploy 2-step pipeline. This will allow us to check builds of any pr including forks, and will also allow deployment of forks. It also maintains safety, assuming the person who approves the workflows checks the code before doing so

nguyentvan7:
## Describe your changes

* Changes described below should allow forks to have proper deployments while also deploying from a correct merge commit, unlike previous deployment implementations.
* Change PR deployment to use artifact-based system
* Build pipelines are triggered on PRs using `pull_request` event
  * This event has no secret access for forks, and requires contributor approval based on the repo settings.
  * This event will also wait to trigger until the mergeability check is complete and there is a merge commit available.
* Once the build is complete, it will archive the output.
* The completion of the build pipeline will trigger a separate deployment pipeline (this will not be shown in the Checks of a PR)
* Deployment pipeline runs from our master branch, and has full secret permissions.
* Deployment pipeline will download the artifact, push the contents to our deployment repo, and spawn/edit a comment on the original PR
* Moved the clean pipeline to use `pull_request_target`. 
  * This event has secret access for forks, and does not require approval.
  * This event always runs the workflow contents from `master` branch, so a developer cannot hijack our pipeline.
  * This event will trigger before the mergeability check is complete.
  * This isn't an issue because we don't need the merge commit for this to run, we just need secret permissions.

## Issue or discord link

- 
- Resolves #1575 

## Testing/validation

Tested on my own repo https://github.com/nguyentvan7/genshin-optimizer/pull/27
Deployment succeeded, comment appeared as expected. We won't be able to test the fork-based capabilities until this is merged into the main repo. But at the very least, it is not a regression for non-fork branches.

### DIFF
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -64,3 +64,4 @@ jobs:
             dist/apps/${{inputs.frontend_name}}
             frontend_name
             deployment_name
+            ref
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -46,7 +46,7 @@ jobs:
         run: |
           printf '%s' "$ENVFILE" > apps/frontend/.env.local
         env:
-          ENVFILE: ${{ secrets.ENVFILE }}
+          ENVFILE: ${{ vars.ENVFILE }}
       - name: Build genshin-optimizer
         run: |
           NX_URL_GITHUB_GO_CURRENT_VERSION="https://github.com/${{ inputs.repo_full_name }}/commit/$(git rev-parse HEAD)" \
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Archive build
         uses: actions/upload-artifact@v4
         with:
-          name: ${{inputs.frontend_name}}
+          name: frontend_build
           path: |
             dist/apps/${{inputs.frontend_name}}
             frontend_name
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -1,0 +1,68 @@
+name: Build Frontend
+run-name: Build frontend of ${{ inputs.frontend_name }} for ${{ inputs.deployment_name }} - ${{ github.event.inputs.ref }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      frontend_name:
+        description: 'Frontend type to build. Can be `frontend`, or `sr-frontend`'
+        type: string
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy-frontend.yml
```diff
@@ -1,117 +1,49 @@
-name: Deploy Frontend
-run-name: Deploy frontend of ${{ inputs.frontend_name }} for ${{ inputs.deployment_name }} - ${{ github.event.inputs.ref }}
+name: Deploy frontend for artifact
```
frzyc, line 1: (nit) would it be "Deploy frontend from built artifact"?

.github/workflows/build-frontend.yml
```diff
@@ -0,0 +1,67 @@
...
+          node-version: 20
+          cache: 'yarn'
+      - name: Yarn install
+        run: |
+          yarn install --immutable --immutable-cache
+      - name: Setup .env
+        run: |
+          printf '%s' "$ENVFILE" > apps/frontend/.env.local
+        env:
+          ENVFILE: ${{ secrets.ENVFILE }}
```
eeeqeee, line 49: We would need to move the `ENVFILE` to vars
```suggestion
          ENVFILE: ${{ vars.ENVFILE }}
```

frzyc (dismissed):
LGTM, looks like it works on your own repo as well. One minor nit.

.github/workflows/build-frontend.yml
nguyentvan7: Good catch

frzyc (approved):

Issue #1548, opened 2024-02-25
Add a toggle to disable using artifacts in teammates' active builds when optimizing


frzyc:
Add a toggle to the optimize tab in team, to disallow using artifacts that are in teammates' builds.

HassnHamada:
I'm working on this issue.

I'll also look into the problem causing negative numbers to appear in "Artifact Configuration" since they are probably in the same module/directory.
<img width="570" alt="Screenshot 2024-03-01 at 11 50 25 AM" src="https://github.com/frzyc/genshin-optimizer/assets/44922829/4fc3954c-6e10-4f6c-95bc-2bb91f6a376b">

frzyc:
Sounds good, let me know if you hit any blockers, this is definitely more complicated logic than tasks you've worked on before.

PR #1602, opened 2024-03-02
Exclude teammates' Builds

HassnHamada:
## Describe your changes
* Solve negative number UI problem
* Add a checkbox to use/ignore teammates loadouts
* Still needs Translation 

## Issue or discord link

Fix: #1548 

## Testing/validation
_1500x866@24


apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/AllowChar.tsx
```diff
@@ -291,7 +291,8 @@ export default function AllowChar({
 
   const onMouseUp = useCallback(() => setMouseUpDetected(true), [])
 
-  const total = locList.length
+  // `total` shouldn't rely on `locList.length` because it gets filtered
+  const total = database.chars.keys.length - 1
```
frzyc, line 295: Travelers are multiple characters in the character database, e.g., `TravelerAnemo` so you would need to convert it to a list of locations, which is what `locList` represents.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ExcludeLoadout.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const database = useDatabase()
+  const {
+    teamCharId,
+    teamChar: { optConfigId },
+    team,
+  } = useContext(TeamCharacterContext)
+  const { useLoadoutExclusion } = useOptConfig(optConfigId)
+
+  useEffect(() => {
+    console.log('Update ExcludeLoadout')
```
frzyc, line 21: console.log

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ExcludeLoadout.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const chars = team.teamCharIds
+      .filter((id) => id !== teamCharId)
+      .map((id) => database.teamChars.get(id))
+    const buildIds = chars.map((char) => char.buildIds).flat()
```
frzyc, line 25: This excludes "every artifact in every build for every character", the feature is intended to exclude "every artifact in the **active** build for every character".

You can approach this via

```
const artIdsToExclude:string[] = team.teamCharIds
      .filter((id) => id !== teamCharId)
      .flatmap((id) => database.teamChars.getLoadoutArtifacts(id))
      .map((art:?ICachedArtifact)=>art?.id)
      .filter(id=>!!id)
 ```

File: apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/index.tsx [hunk 4]
```diff
@@ -194,10 +195,15 @@ export default function TabBuild() {
...
       allowLocationsState,
       useExcludedArts,
+      useLoadoutExclusion,
+      loadoutExclusion,
     } = deferredArtsDirty && deferredBuildSetting
 
+    console.log('Run Filter')
     return database.arts.values.filter((art) => {
       if (!useExcludedArts && artExclusion.includes(art.id)) return false
+      if (!useLoadoutExclusion && loadoutExclusion.includes(art.id))
```
frzyc, line 205: This should be where you apply the logic in the `useEffect` above.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/index.tsx
```diff
@@ -194,10 +195,15 @@ export default function TabBuild() {
       levelHigh,
       allowLocationsState,
       useExcludedArts,
+      useLoadoutExclusion,
+      loadoutExclusion,
     } = deferredArtsDirty && deferredBuildSetting
 
+    console.log('Run Filter')
```
frzyc, line 202: console.log

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ExcludeLoadout.tsx
```diff
@@ -0,0 +1,55 @@
...
+      color={useLoadoutExclusion ? 'success' : 'secondary'}
+      onClick={() => {
+        database.optConfigs.set(optConfigId, {
+          useLoadoutExclusion: !useLoadoutExclusion,
+        })
+      }}
+      disabled={disabled}
+    >
+      {/* TODO: Translation */}
+      Use Teammates Loadouts
```
frzyc, line 52: `Use artifacts in teammates' builds`

frzyc (changes requested):
Some concerns with implementation and system design.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ExcludeLoadout.tsx
frzyc: admittedly, I should name this function to `getActiveBuildArtifacts` to be more appropriate.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/AllowChar.tsx
HassnHamada: I don't understand the problem you are trying to say. But doesn't the `database.chars` stores every character I have in my `/characters` tab. So it's length should be the total count of characters I have.

-----
The problem with `locList` is that with the current implementation it  changes based on the applied filters Which is causing the problem of negative numbers.

-----
When starting everything looks good
Screenshot 2024-03-03 at 8 50 44 PM

-----
Then maybe someone wants to remove all characters. which causes the `excludedLocations.length` to increase and `useTot` to be `0` which is right.
Screenshot 2024-03-03 at 8 51 21 PM

-----
Until they decide "nah maybe I should add a certain character" so they use filtering at this point `locList` get reduced to the number to the remaining characters while `excludedLocations` remains the same causing `useTot` to be negative.
Screenshot 2024-03-03 at 8 51 27 PM
Screenshot 2024-03-03 at 8 51 41 PM

-----
Another solution may be to change the `onClose` callback to clear the filters so `locList` size would be reset to origin.

frzyc: I understand this might be confusing. my concern was mainly due to the fact that travelers in GO share artifacts, so when you say "take from TravelerAnemo"(characterkey),  you are really saying "take from traveler"(location key). So when you have a tally of "characters" you can take from, the tally is really counting the "number of locations to take artifacts from"

HassnHamada: Ok. I Got It :/

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/index.tsx
HassnHamada: Can you explain more what you mean!

frzyc: sure, the code here should look like:
```
const artsOnTeammates = teammates.flatmap(tm=>db.getLoadoutArtifacts.id)


return database.arts.values.filter(art=>{
...
if(!useTeammatesBuild && artsOnTeammates.includes(art.id)) return false
...
})

```

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -85,6 +85,8 @@ export interface OptConfig {
   compareBuild: boolean
   levelLow: number
   levelHigh: number
+  useLoadoutExclusion: boolean // teammates loadout exclusion flag
+  loadoutExclusion: string[]
```
frzyc, line 89: I don't think its a good idea to store this value in the database, because this value is calculatable.

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -85,6 +85,8 @@ export interface OptConfig {
   compareBuild: boolean
   levelLow: number
   levelHigh: number
+  useLoadoutExclusion: boolean // teammates loadout exclusion flag
```
frzyc, line 88: I would rename this to `useTeammateBuild`. We want to exclude only the active build on teammates, not their entire loadout.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ExcludeLoadout.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const database = useDatabase()
+  const {
+    teamCharId,
+    teamChar: { optConfigId },
+    team,
+  } = useContext(TeamCharacterContext)
+  const { useLoadoutExclusion } = useOptConfig(optConfigId)!
+
+  useEffect(() => {
+    const artifactIds = Array.from(
```
frzyc, line 21: Move logic here into the filtering code called on optimize, since you want to recalculate teammates artifacts per optimization.

frzyc (changes requested):

frzyc (changes requested):
minor changes to make variables names better

frzyc (approved):
LGTM, thanks for sticking through with this feature!

HassnHamada:
NOOOOO, wait

PR #1608, opened 2024-03-03
Fix active character usage and tc opt

nguyentvan7:
## Describe your changes

* Add util function to get the active character index in a team (in case this changes later or we need to refactor for multiple active)
* Adjust which character is marked as active when optimizing (tc + upopt + regular)
* Fix tc opt to look at the actual teams data, not just the singular character

## Issue or discord link

- 

## Testing/validation

On current master, I did a TC solve for Ganyu with:
* Rosaria providing 15 CR
* Hunters Path
* ATK/Cryo/CD mainstats
* Blizz 4p on Frozen
* KQMS
And the solver still tried to give me 12 CR rolls for 145 total CR. So seems like it was ignoring all the conditional CR.

On this branch, I did the same thing and got a 103 CR build, with only the 2 base rolls of CR. Which mathematically makes more sense.

### DIFF
--- a/libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
+++ b/libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
@@ -87,12 +87,9 @@ export class TeamDataManager extends DataManager<
     return id
   }
 
-  getActiveTeamCharId(teamId: string) {
-    const team = this.database.teams.get(teamId)
-    return team?.teamCharIds[0]
-  }
   getActiveTeamChar(teamId: string) {
-    const teamCharId = this.getActiveTeamCharId(teamId)
+    const team = this.database.teams.get(teamId)
+    const teamCharId = team?.teamCharIds[0]
     return this.database.teamChars.get(teamCharId)
   }
 }
--- a/libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
+++ b/libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
@@ -119,3 +119,7 @@ function validateTeam(
     lastEdit,
   }
 }
+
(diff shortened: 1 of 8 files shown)


libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
```diff
@@ -119,3 +119,7 @@ function validateTeam(
     lastEdit,
   }
 }
+
+export function getActiveTeamCharId(team: Team) {
```
frzyc, line 123: Can you attach this function to `TeamDataManager`? so its usage should be `database.teams.getActiveTeamCharId(teamId)`

frzyc (changes requested):
Minor change to db code

libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
nguyentvan7: In the places this function is called, the `team` object is already fetched. Isn't your suggestion requiring a redundant fetch of data?

frzyc: Yes, but I think standardizing the APIs is more important than a single fetch (which is just a lookup)

frzyc (dismissed):
Looks to be working

libs/gi/db/src/Database/DataManagers/TeamDataManager.ts
nguyentvan7: Done

nguyentvan7:
Updated description with how I tested

frzyc (changes requested):
format

frzyc (approved):

PR #1597, opened 2024-03-01
Allow creating a separate build when equipping a build to currently equipped

kasantos2:
## Describe your changes

When equipping a build, show an improved confirmation modal and allow users to create a new build for the current equipment.

## Issue or discord link

- Closes #1547 

## Testing/validation

gi-1547

## Checklist before requesting a review (leave this PR as draft if any part of this list is not done.)

- [ ] I have commented my code in hard-to understand areas.
- [ ] I have made corresponding changes to README or wiki.
- [ ] For front-end changes, I have updated the corresponding English translations.
- [x] I have run `yarn run mini-ci` locally to validate format and lint.
- [ ] If I have added a new library or app, I have updated the deployment scripts to ignore changes as needed

PR #1618, opened 2024-03-04
Add modal to compare weapons in generated builds in the optimize tab and disable individual swapping/locations

### LINKED ISSUE #1591: Optimized build results have no way to compare weapon
Due to the resolution of https://github.com/frzyc/genshin-optimizer/issues/1565, it is now possible to have a build result have a different weapon than the weapon in the currently equipped build.

image
In this image, I have a different weapon in my comparison build, than what is being optimized for. But I have no way to compare the 2 weapons. When I equip, it will swap my weapon over.

It would be nice to have a modal similar to the artifact swap modal
image

### LINKED ISSUE #1616: Disable individual artifact swapping in `CompareArtifactModal`
Since an optimal build for a set of criteria is defined by a combination of weapon and a specific set of artifacts, it doesn't make a whole lot of sense to cherry-pick individual artifacts from the generated builds over saving/equipping the whole build. Artifact swapping in this UI should be disabled.
image

The rest of the UI is still useful for artifact inclusion/exclusion and management (i.e. generated build uses an artifact that the user no longer has) as well as comparing artifact differences between a generated build and a currently inactive build and should otherwise remain intact. However, equipping functionality should be handled entirely by the "Equip Build" button.

arquelion:
## Describe your changes
- Added `CompareWeaponModal` to `BuildDisplayItem.tsx` that functions identically to the existing `CompareArtifactModal`
- Removed the ability to swap artifacts from `CompareArtifactModal`
- Disabled the location dropdowns for both comparison modals
- Fixed a bug where the old weapon/artifact shown in the comparison modal does not reflect any changes made to the comparison build despite calcs updating in real-time

## Issue or discord link
- Resolves #1591 
- Resolves #1616
- https://discord.com/channels/785153694478893126/1200154094476734556/1214119792999403591

## Testing/validation
Comparison weapon matches optimized weapon
image

Comparison weapon does not match optimized weapon
image

Modified artifact comparison
image

### DIFF
--- a/apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
+++ b/apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
@@ -18,6 +18,7 @@ import {
   useArtifact,
   useDatabase,
   useOptConfig,
+  useWeapon,
 } from '@genshin-optimizer/gi/db-ui'
 import { Checkroom, ChevronRight } from '@mui/icons-material'
 import CheckBoxIcon from '@mui/icons-material/CheckBox'
@@ -48,6 +49,7 @@ import { uiInput as input } from '../../../../../Formula'
 import ArtifactCard from '../../../../../PageArtifact/ArtifactCard'
 import { ArtifactSetBadges } from './ArtifactSetBadges'
 import SetInclusionButton from './SetInclusionButton'
+import WeaponCard from '../../../../../PageWeapon/WeaponCard'
 
 type NewOld = {
   newId: string
@@ -87,8 +89,10 @@ export default function BuildDisplayItem({
   ) ?? { mainStatAssumptionLevel: 0, allowLocationsState: 'all' }
   const database = useDatabase()
   const { data } = useContext(DataContext)
-  const [newOld, setNewOld] = useState(undefined as NewOld | undefined)
-  const close = useCallback(() => setNewOld(undefined), [setNewOld])
+  const [weapNewOld, setWeapNewOld] = useState(undefined as NewOld | undefined)
+  const [artNewOld, setArtNewOld] = useState(undefined as NewOld | undefined)
+  const closeWeap = useCallback(() => setWeapNewOld(undefined), [setWeapNewOld])
+  const closeArt = useCallback(() => setArtNewOld(undefined), [setArtNewOld])
   const buildEquip = buildId && buildType === 'real'
   const equipBuild = useCallback(() => {
     const confirmMsg = buildEquip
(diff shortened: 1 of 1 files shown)


frzyc (changes requested):
Your changes here seems to revert changes added in https://github.com/frzyc/genshin-optimizer/pull/1597, can you see if its an issue with your git history?

arquelion:
Looks like the revert I did after accidentally merging in master when I meant to update my own master branch undid those changes. Should be fixed now.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -87,8 +88,10 @@ export default function BuildDisplayItem({
   ) ?? { mainStatAssumptionLevel: 0, allowLocationsState: 'all' }
   const database = useDatabase()
   const { data } = useContext(DataContext)
-  const [newOld, setNewOld] = useState(undefined as NewOld | undefined)
-  const close = useCallback(() => setNewOld(undefined), [setNewOld])
+  const [weapNewOld, setWeapNewOld] = useState(undefined as NewOld | undefined)
```
frzyc, line 91:
```Typescript

// update when data is recalc'd
const weaponNewOld = useMemo(()=>({
  oldId: oldData.get(input.weapon.id).value!
  newId: data.get(input.weapon.id).value!
}),[data,oldData])

//state for showing weaponmodal
const [showWeapon, onShowWeapon, onHideWeapon] = useboolean()
```

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -87,8 +88,10 @@ export default function BuildDisplayItem({
...
+  const [artNewOld, setArtNewOld] = useState(undefined as NewOld | undefined)
```
frzyc, line 92:
```Typescript
// update when data is recalc'd
const artNewOldBySlot = useMemo(()=>
  objKeyMap(allArtifactSlots,(slotKey)=>({
    oldId: oldData.get(input.art[slotKey].id).value ??''
    newId: data.get(input.art[slotKey].id).value ??''  
  )) as Record<SlotKey,NewOld>
,[data,oldData])

//state for showing art compare modal
const [artNewOld, setArtNewOld] = useState<NewOld|undefined>()
```

frzyc (changes requested):
I just noticed that there is a bug, where users can change the artifacts/weapon on the "compare build", and the calculation comparison on generated builds would recalculate, but the popup for comparison (new/old) of art/weapon would not be updated. 
There are two ways we can fix this. 
1. register for updates with the compared build, which can trigger UI updates,
2. Take the weapon/artifact ids from the calculated Data. 

I think the 2nd approach is more "elegant", since a recalculation data is already propagated, we can piggy back off it without adding additional listeners. I added some pseudocode to get you started.

arquelion:
Fixed, I've manually checked it locally but can add validation images/recordings later.

Looks like this bug was already present in top of tree `master` so I've also added the fix to the PR description

frzyc (approved):

frzyc:
LGTM, thanks for addressing the fix as well.

PR #1622, opened 2024-03-05
Cache more actions

eeeqeee:
## Issue or discord link

- 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -92,8 +93,13 @@ jobs:
             submodule-
       - uses: actions/checkout@v4
         with:
+          ref: ${{ github.sha }}
```
nguyentvan7, line 96: What does this do? I didn't even see the ref get logged, but it looks like it still just checks out the merge commit

eeeqeee: Currently, when running CI on pushes to master we checkout the repo's master branch as `master` which causes the format check to effectively be skipped. Switching to `github.sha` checks it out as a detached HEAD so format is run on the whole repo instead

nguyentvan7: Could you add a comment then, i forgot this runs on master too

nguyentvan7 (approved):

Issue #1552, opened 2024-02-26
Add a view on character UI to show all teams


frzyc:
rough idea
![image](https://github.com/frzyc/genshin-optimizer/assets/1754901/7f8db09e-936b-4ace-bdc5-9826e24ab910)
Probably not a modal, maybe a separate tab in the character UI.
Shows builds, and nested builds in teams. 

Questions to answer:
- how much interactions do we have here? does clicking on the build go to the build in the team? Does the build buttons do anything (I rather people do detailed build work in the team UI to keep character UI clean)
- If builds are for specific stat breakpoints, do we calculate each team for those stats?
- What does showing all builds do? what does user get out of this information?

HassnHamada:
I was thinking about this and I'll start working on it. I was thinking about removing these two section since the way the GO is implemented now a single character can have multiple build/loadouts based on the team.

<img width="1279" alt="Screenshot 2024-03-06 at 7 13 15 PM" src="https://github.com/frzyc/genshin-optimizer/assets/44922829/0090061b-ee8f-4a73-8cf7-4683c7f941c4">

PR #1628, opened 2024-03-06
Experimenting with character card layout

HassnHamada:
## Describe your changes
Add the currently equipped build to each team

## Issue or discord link

#1552

## Testing/validation

1200x693@18

### DIFF
--- a/apps/frontend/src/app/Components/Character/CharacterEditor/CharacterLoadout.tsx
+++ b/apps/frontend/src/app/Components/Character/CharacterEditor/CharacterLoadout.tsx
@@ -4,9 +4,10 @@ import {
   SqBadge,
 } from '@genshin-optimizer/common/ui'
 import { unit } from '@genshin-optimizer/common/util'
-import { artifactAsset, weaponAsset } from '@genshin-optimizer/gi/assets'
+import { artifactAsset } from '@genshin-optimizer/gi/assets'
 import type { ArtifactSlotKey } from '@genshin-optimizer/gi/consts'
 import { allArtifactSlotKeys } from '@genshin-optimizer/gi/consts'
+import type { ICachedWeapon} from '@genshin-optimizer/gi/db';
 import { type ICachedArtifact } from '@genshin-optimizer/gi/db'
 import { useBuildTc, useDatabase } from '@genshin-optimizer/gi/db-ui'
 import { StatIcon } from '@genshin-optimizer/gi/svgicons'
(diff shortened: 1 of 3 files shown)


frzyc:
This is a very interesting and cool idea, I like the fact that its not inherently taking up a ton of extra space(until its expanded).
Some first thoughts(not demanding an answer, just more noting down thoughts):
- The expanded equipment(2x3) might be a tad big. We can probably reuse the layout from the character card (1x6)
- What happens if the "Active build" is a TC build?
- This creates more of an association of the build with the team, when we want to associate the build with the loadout(since builds are stored per loadout). in a loadout, all the teams actually have the same "active build", since they are using the same loadout.

HassnHamada:
* I just tried it and it defently looks cleaner
* I'm using `TeamCharacterDataManager.getLoadoutArtifacts` and `TeamCharacterDataManager.getLoadoutWeapon` currently they return nothing if the active build is a TCBuild I'll look into them image
* Yes It makes so much more sense to have it per loadout instead of per team. My idea is to move it on the top of the teams as a separate card

HassnHamada:
## TC Build
Screenshot 2024-03-08 at 1 13 31 AM

## Normal Build
Screenshot 2024-03-08 at 1 15 36 AM

I'm think of removing all the hovering cards since they always show the "equipped" not what is actually active. And I'll also think of a way show the artifacts in batter "smaller" style 
Screenshot 2024-03-08 at 1 15 44 AM

HassnHamada:
@frzyc How do I get the status of the current loadout? I can do it manually by adding the status of each piece of artefact, the set effect, and the weapon. I'm just asking if there is an easier way, or should I just remove the status section? 

Screenshot 2024-03-10 at 10 08 17 PM
Screenshot 2024-03-10 at 10 08 09 PM

frzyc:
@HassnHamada For single character only calculations, you can use `useCharData` and feed it the build via `overrideArt` and `overrideWeapon`. We prefer to do genshin calculations using the engine itself, rather than calculating it manually on the UI side.

HassnHamada:
@frzyc I don't have any more plans for this if you think there is any UI changes that need to be done or bad implementations just inform me I'll work on it 💪

frzyc:
I think you make a good point about how the teamcard does not show the "loadout character" but the "equipped". In the character editor, I think you are right to disable the hover.

In the teams UI, I am going to make the hover show the loadout name + active build on that loadout.

frzyc:
I appreciate all the effort you went into iterating on the design, and you definitely presented a lot of possibility towards evolving UI here, but I think ultimately we need to keep the focus of this UI to what's important(showing loadouts and associating them with teams). I think we can leverage some of your UI ideas to other parts of the UI(especially parts that show builds, maybe)

Issue #1606, opened 2024-03-03
Bring "Team Editor" outside of "Team Settings" popup


StainAE86:
Currently the character selector strip across the top is redundant. Since you can still swap to the different characters within the team via the Team Buffs tab like we've always done. So adding this new selector strip/tabs in the loadout feature without some additional function just adds clutter to the interface.

Why not allow people to add/remove/order characters via this selector strip?
This then keeps the "Team Settings" and "Loadout Settings" popups consistent with eachother - existing to just name/describle the contents.
![image](https://github.com/frzyc/genshin-optimizer/assets/101696552/ce2f5f62-d1be-4435-a5b1-8e1c8e0f2333)

Here is an idea for the proposed layout: (forgot to add a button to represent swapping/changing a team mate)
![image](https://github.com/frzyc/genshin-optimizer/assets/101696552/62667c0d-45fd-438f-9c57-7d0ff374e097)

But perhaps use nicer UI elements like these within the character strip:
![image](https://github.com/frzyc/genshin-optimizer/assets/101696552/2addf1c8-4cc4-4468-9204-7aa4ec7d9576)

This also includes the loadout selection dropdown directly inside the Team Buffs tab above the associated character so you can see which "version" of that character you have selected - down to what conditional buffs they're providing and can quickly swap between them once you've realised you've got the wrong version of them.

StainAE86:
Different approach to this would be folding the "team settings" and "team buffs" into 1 new "Team" tab. (leaving the top character selector strip as-is)

![image](https://github.com/frzyc/genshin-optimizer/assets/101696552/3fbac626-a93a-4f54-93c0-d29087b6bcf3)
Removing the team buff stats allows for a 4th column to put all characters in this new "Team" tab, that now gives the ability to select/remove characters from each slot, and select their loadout. (the functions of the "team settings" popup)
Character lvl/talents/cons are not really relevent here so can be removed to tidy the interface.

Also if 4 characters are not present, can remove the placeholder(s) from the character selector strip.

The removed team buff stats would then be present on the optimize tab where they are more relevent.
![image](https://github.com/frzyc/genshin-optimizer/assets/101696552/d2db5939-865d-4ae7-9115-b07c1f6ac0ae)

![image](https://github.com/frzyc/genshin-optimizer/assets/101696552/397be74f-c66b-45a8-bebf-ec9d889e72fc)
And is also where the "bonus stats" appear anyway. So it makes it consistent.

[...]

PR #1632, opened 2024-03-08
move teambuff to team overview

frzyc:
## Describe your changes
Move the content of the teambuff tab to the team setting modal, which now serves as a "team overview"
image
Since this is the entry point towards engaging with the teams section, users are aware of teambuffs that interact with loadouts, and at a glance, users can also get a general idea if their loadouts are correct.

 - Resolves https://github.com/frzyc/genshin-optimizer/issues/1606

### DIFF
--- a/apps/frontend/src/app/App.tsx
+++ b/apps/frontend/src/app/App.tsx
@@ -146,7 +146,7 @@ function Content() {
       <Grid item>
         <Header anchor="back-to-top-anchor" />
       </Grid>
-      <Container maxWidth="xl" sx={{ px: { xs: 0.5, sm: 1, md: 2 } }}>
+      <Container maxWidth="xl" sx={{ px: { xs: 0.5, sm: 1 } }}>
         <Suspense
           fallback={
             <Skeleton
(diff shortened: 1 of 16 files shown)


nguyentvan7:
Why was the outer card removed for teammates? The floating effect looks nice/new on desktop, but it makes the UI really hard to parse on mobile
Screenshot_20240308_075257_Chrome

frzyc:
> Why was the outer card removed for teammates? The floating effect looks nice/new on desktop, but it makes the UI really hard to parse on mobile 

Because there are now 4 characters in a 5 column layout, I thought the character card was using up too much padding. There is not much we can do with mobile other than creating separate UI just for mobile, which is something I don't really want to undertake until the UI around this area hits maturity (don't want to spend time on mobile layout optimization when we havent finalized the desktop UI)

nguyentvan7:
Also I really dislike the tall Add Character button. Can we make it a fixed height instead of 100% height

StainAE86:
> Also I really dislike the tall Add Character button. Can we make it a fixed height instead of 100% height

Agreed

nguyentvan7 (approved):
Still ambivalent but approving to unblock

Issue #1592, opened 2024-03-01
Adjust wording of "Currently Equipped" in optimized build


nguyentvan7:
![image](https://github.com/frzyc/genshin-optimizer/assets/36019388/1d0d5e25-aed4-4ad6-9a8f-ef572ebc92e3)
This should maybe say something like "Current Build"

PR #1657, opened 2024-03-11
Add indicator for current build/equipped in generated builds

frzyc:
## Describe your changes

Show whether a generated build is equipped (and if its a saved build)
image

Show whether a generated build is the "current build" (and if multiple builds match)
image

Update Opt Graph
image

## Issue or discord link

- Resolves https://github.com/frzyc/genshin-optimizer/issues/1592

### DIFF
--- a/apps/frontend/src/app/PageTeam/CharacterDisplay/Build/BuildEquipped.tsx
+++ b/apps/frontend/src/app/PageTeam/CharacterDisplay/Build/BuildEquipped.tsx
@@ -49,7 +49,6 @@ export function BuildEquipped({ active = false }: { active: boolean }) {
     <CardThemed
       bgt="light"
       sx={{
-        undefined,
         boxShadow: active ? '0px 0px 0px 2px green inset' : undefined,
       }}
     >
--- a/apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ArtifactSetBadges.tsx
+++ b/apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/ArtifactSetBadges.tsx
@@ -13,25 +13,23 @@ import { getArtSheet } from '../../../../../Data/Artifacts'
 
 type ArtifactSetBadgesProps = {
   artifacts: ICachedArtifact[]
-  currentlyEquipped?: boolean
 }
-export function ArtifactSetBadges({
-  artifacts,
-  currentlyEquipped = false,
-}: ArtifactSetBadgesProps) {
-  const setToSlots: Partial<Record<ArtifactSetKey, ArtifactSlotKey[]>> =
-    useMemo(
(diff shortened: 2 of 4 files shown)


arquelion:
Since the current behavior of the "Equip Build" button is to equip to the currently active build (which may not necessarily be the same as the equipped build), shouldn't the button be enabled in the first image and disabled in the second instead?

(I hope you don't mind my sticking my nose in. I've been working on adding the full swap preview when equipping over an existing build and ran into a related issue with the "Equip Build" button, popped in to see if you'd happened to address this already as part of this PR.)

frzyc:
> Since the current behavior of the "Equip Build" button is to equip to the currently active build (which may not necessarily be the same as the equipped build), shouldn't the button be enabled in the first image and disabled in the second instead?
> 
> (I hope you don't mind my sticking my nose in. I've been working on adding the full swap preview when equipping over an existing build and ran into a related issue with the "Equip Build" button, popped in to see if you'd happened to address this already as part of this PR.)

You are completely right. this is a mistake on my side.

arquelion:
I forgot to open an official issue for it since I was trying to dig for a solution, but it was almost identical now that I think about it. The button would remain enabled after equipping to the currently active build when the active build wasn't the same as the equipped build until after I refreshed the page.

I do have a closely related issue where my preview of the currently active build won't update either until after the refresh, something about the old `buildEquippedArtifactIds` and `buildEquippedWeaponId` values not being re-computed despite the changes to the database from `equipBuild`. Hopefully this will also be indirectly resolved in the process.

frzyc:
> I forgot to open an official issue for it since I was trying to dig for a solution, but it was almost identical now that I think about it. The button would remain enabled after equipping to the currently active build when the active build wasn't the same as the equipped build until after I refreshed the page.
> 
> I do have a closely related issue where my preview of the currently active build won't update either until after the refresh, something about the old `buildEquippedArtifactIds` and `buildEquippedWeaponId` values not being re-computed despite the changes to the database from `equipBuild`. Hopefully this will also be indirectly resolved in the process.

Seems like an update issue, I'd say in situation like this, always defer to creating an issue, it will be more visible.

nguyentvan7:
image
i currently have this build Equipped, and I am on my Equipped build. but it is saying that "vsva" is my current build, which doesn't seem right

nguyentvan7:
image

I have this build as both "bkj" and "kbkj" saved. but it only shows as "bkj". I also think just inserting the build name here is a little confusing. What if we changed the badge to be a badge with a constant label, then when you hover, it shows you all the builds that it matches?

nguyentvan7 (commented):
see previous comments

frzyc:
> see previous comments

Forgot to push the code. Basically did what you suggested, have a popup that shows the matching builds, if there are multiple.

nguyentvan7 (approved):

Issue #1683, opened 2024-03-13
Comparison easily gets cramped in smaller layout


nguyentvan7:
Similar to #1504 except much easier to trigger
![image](https://github.com/frzyc/genshin-optimizer/assets/36019388/c7c88d8e-5362-43ce-b30a-aa2745ae2ea4)

In this breakpoint, the stats are 3 column, but the character card column is taking up so much space that it is squeezing the 3 columns. In this case, it is super easy to overflow a column even with just small number changes for the comparison. 

We should prioritize fixing this breakpoint due to the ease at which it breaks.

nguyentvan7:
![image](https://github.com/frzyc/genshin-optimizer/assets/36019388/616f4d46-d31f-42dd-8f3f-d9b2bcfe6e9e)
Here is the next breakpoint. the character card is significantly smaller. so this can probably be a good first step to reducing the cramped-ness

PR #1686, opened 2024-03-13
Reduce columns for stat display

frzyc:
## Describe your changes

Since the comparison UI were still scaled before we had % comparison, the increase in content due to comparison is pushing content off the grid.

## Issue or discord link

- Resolves https://github.com/frzyc/genshin-optimizer/issues/1683


nguyentvan7:
The single stats column makes it suuuuper long. Could we try reducing the width of the character card/edit button instead? They really don't need to take up like 40% of the width

nguyentvan7:
I think with the character card shrinkage, we can still have 2-column smallest breakpoint for stats

frzyc:
> I think with the character card shrinkage, we can still have 2-column smallest breakpoint for stats

There is no way 2 columns can fit on the smallest breakpoint. it is smaller than 600px, with a reasonable support of 360px.
image
image

nguyentvan7 (approved):

PR #1733, opened 2024-03-15
Allow invisible backdrop click-through

### LINKED ISSUE #1497: Create a custom Dialog override to prevent overlay race condition
Similar to #967, we need to modify Dialog to use Popper instead of Popover or something similar. The backdrop is getting stuck on certain race conditions which renders the whole site unusable.

Probably related to https://github.com/mui/material-ui/issues/32286

frzyc:
## Issue or discord link

- Resolves #1497

### DIFF
--- a/libs/common/ui/src/theme/index.ts
+++ b/libs/common/ui/src/theme/index.ts
@@ -215,5 +215,18 @@ export const theme = createTheme({
         },
       },
     },
+    /*
+     * FIXME: A ducttape solution to address
+     * https://github.com/mui/material-ui/issues/32286
+     */
+    MuiBackdrop: {
+      styleOverrides: {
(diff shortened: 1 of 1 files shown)


nguyentvan7:
Doesn't work

nguyentvan7:
Here is the blocking element 
image

nguyentvan7:
If you're willing to investigate more, I definitely think its related to the re-render cycle caused by applying KQMS

frzyc:
> If you're willing to investigate more, I definitely think its related to the re-render cycle caused by applying KQMS

I don't think it's that simple, since I did add a timout with applying the KQMS as one experiment, with varying success. Its probably not worth the time to investigate.

PR #1759, opened 2024-03-16
show team buff per character

### LINKED ISSUE #1755: Show teambuffs per character in the team setting UI
Problem statement:
As we make more distinctions between teambuffs in terms of the target(s) that receives the buff (on-field, full team, full team without src, specific element), a singular teambuff UI is not sufficient. 

The current Team buff UI actually reflects the current character in team UI's teambuffs, but since there is no indicator that is what's happening, its not intuitive. 

Suggestion:
In the team setting UI, duplicate the teambuff UI across all teammates. 
image

Questions to resolve:
- Maybe we can move the resonance UI out of the 1st column, so we can have a "4-column" layout instead of 5.

frzyc:
## Describe your changes

Show Teambuff per character to better show non-teamwide buffs between teammembers

## Issue or discord link

- Resolves #1755 

## Testing/validation

On a geo resonance team, can see the geo resonance and other on-field buffs are only applied to the first character.
image

### DIFF
--- a/apps/frontend/src/app/PageTeam/TeamSetting/TeamComponents.tsx
+++ b/apps/frontend/src/app/PageTeam/TeamSetting/TeamComponents.tsx
@@ -213,6 +213,7 @@ export function TeammateDisplay({
             <CharacterCardWeaponFull loadoutDatum={loadoutDatum} />
             <CharWeaponCondDisplay />
             <CharTalentCondDisplay />
+            <TeamBuffDisplay />
           </Suspense>
         </CharacterContext.Provider>
       </DataContext.Provider>
--- a/apps/frontend/src/app/PageTeam/TeamSetting/TeamComponents.tsx
+++ b/apps/frontend/src/app/PageTeam/TeamSetting/TeamComponents.tsx
@@ -68,7 +68,7 @@ export function TeamBuffDisplay() {
   return (
     <CardLight>
       <CardContent>
-        <Typography>Team Buffs</Typography>
+        <Typography>Received Team Buffs</Typography>
       </CardContent>
       <Divider />
       <CardContent>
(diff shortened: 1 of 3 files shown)


nguyentvan7:
Should this be at the top? and maybe collapsed by default for each teammate? Its a bit noisy, and not having them on the same y-level makes it hard to compare

frzyc:
> Should this be at the top? and maybe collapsed by default for each teammate? Its a bit noisy, and not having them on the same y-level makes it hard to compare

I'm not inclined to fiddle around with the UI too much, I just want to get this in to fix a gap, and we can polish this UI later

nguyentvan7:
I don't think it's really in a state to be released

frzyc:
its showing information that was not accessible before, and due to that, it should be released to address the gap. Tidying it up can be done later when we actually have time for a discussion on what to do with this UI (we've already spent a bit of time yesterday with no conclusion, so i doubt it will be quick)

StainAE86:
> its showing information that was not accessible before, and due to that, it should be released to address the gap. (I'd argue this is a "fix ASAP" prio) Tidying it up can be done later when we actually have time for a discussion on what to do with this UI (we've already spent a bit of time yesterday with no conclusion, so i doubt it will be quick)

I don't think it's communicated well enough that these are the buffs that the above character is recieving.
And I think it would look neater if this overview of buffs were above the conditionals rather than the last item in the column.
It's visually very messy when there's very little horizontal alignment - makes it hard to parse.

nguyentvan7:
I don't think the gap is so urgent we need to rush this out the door. The changes im asking are like 5 mins at max. Also we were iterating over the design of team buffs in general yesterday, this is a much smaller scoped issue

frzyc:
I'm fine with reordering it higher, but its a lot of extra work to make them "line up", the closest is that they are in the same general horizontal area.

nguyentvan7 (approved):
Good enough

Issue #1756, opened 2024-03-16
Clarify Team header button functions


frzyc:
Currently there are quite a lot of things residing within the header of the teams page.
![image](https://github.com/frzyc/genshin-optimizer/assets/1754901/73ad0934-f81e-4353-af56-9100a2a0f3fd)
The two more important buttons, team settings and loadout settings, are not quite evident, and other features like mtarget, bonus stats, are not very easy to see.

Suggestions to address, by reorganizing the rows: (where [] denotes buttons)
1. [Team name]
2. teammate selector
3. [Loadout name | Build name]
4. [Bonus Stats] [Multi opt] [ formulas & calcs]

hitmode toggles can be moved closer to stat display, other toggles can be removed:

- https://github.com/frzyc/genshin-optimizer/issues/1646
- https://github.com/frzyc/genshin-optimizer/issues/1645

Questions:
- should enemy config be folded into the team settings?
- To reduce the color noise, perhaps we can use outlined buttons? (make them the same color as the element of the current character?)

PR #1768, opened 2024-03-17
Revamp team UI header

frzyc:
## Describe your changes
The buttons on the header of the team UI seems to be very loud, and there is not a visual distinction between "team actions" and "character actions". 

- Converting header buttons to outline buttons (and provide a lighter background so show button contrast still).
- Use the character's element for character specific buttons.
- Loadout/build setting button use the full width, so longer names would not cause too much layout shift.
- Character tabs selector to use the character's background.
- lighten the hydro + geo color to increase contrast.

## Issue or discord link

- Resolves https://github.com/frzyc/genshin-optimizer/issues/1756
## Testing/validation

image
image


nguyentvan7 (commented):
I don't like bar.png and colored buttons. I'm ok to merge without those changes, and consider bar.png and colored buttons another time

frzyc:
@nguyentvan7 How about this? lets hit a compromise. I will remove the bar.png but keep the colored buttons.

nguyentvan7:
I still don't like colored buttons. they dont hint anything imo and just create noise. We don't have this design pattern anywhere else that color -> character. this + bar.png just seem like design flair, but they make the UI harder to parse

frzyc:
> I still don't like colored buttons. they dont hint anything imo and just create noise. We don't have this design pattern anywhere else that color -> character. this + bar.png just seem like design flair, but they make the UI harder to parse

Reverting them back to the cyan buttons will be more noisy, since it would be

Elemental color team selector
CYAN
CYAN
elemental tab selector
image

I don't agree with the whole "don't hint anything", since it doesn't really say why you don't like it, other than a denial towards my reasoning. 

I don't think the "we don't have this design pattern anywhere else" is not a good reason to block this either, since we should not be limited by arbitrary "established convention" when we can be constantly trying to challenge it as an evolving product, and we can't have "established convention" without at least introducing it somewhere.

nguyentvan7 (approved):

nguyentvan7 (approved):

PR #1819, opened 2024-03-23
Reduce lag in main stat filter

eeeqeee:
## Describe your changes

Reduce lag when toggling main stat filters and many builds are generated

## Issue or discord link

- https://discord.com/channels/785153694478893126/1220714512512974858

### DIFF
--- a/apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
+++ b/apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
@@ -73,6 +73,8 @@ type BuildDisplayItemProps = {
   disabled?: boolean
   extraButtonsRight?: JSX.Element
   extraButtonsLeft?: JSX.Element
+  mainStatAssumptionLevel: number
+  allowLocationsState: AllowLocationsState
 }
 
 // TODO: Translation for build UI
@@ -82,17 +84,16 @@ export default function BuildDisplayItem({
   extraButtonsRight,
   extraButtonsLeft,
   disabled,
+  mainStatAssumptionLevel,
+  allowLocationsState,
 }: BuildDisplayItemProps) {
   const {
     loadoutDatum: { buildType, buildId },
-    teamChar: { optConfigId, buildIds = [] },
+    teamChar: { buildIds = [] },
   } = useContext(TeamCharacterContext)
   const {
     character: { key: characterKey, equippedArtifacts, equippedWeapon },
   } = useContext(CharacterContext)
-  const { mainStatAssumptionLevel, allowLocationsState } = useOptConfig(
-    optConfigId
-  ) ?? { mainStatAssumptionLevel: 0, allowLocationsState: 'all' }
   const database = useDatabase()
   const { data, oldData } = useContext(DataContext)
 
--- a/libs/common/database/src/lib/DataManagerBase.ts
+++ b/libs/common/database/src/lib/DataManagerBase.ts
@@ -78,7 +78,10 @@ export class DataManagerBase<
       | ((v: StorageValue) => Partial<StorageValue> | void),
     notify = true
   ): boolean {
-    const old = this.getStorage(key)
+    const old =
+      this.dataKey === 'optConfigs'
+        ? this.get(key) ?? this.getStorage(key)
+        : this.getStorage(key)
     if (typeof valueOrFunc === 'function' && !old) {
       this.trigger(key, 'invalid', valueOrFunc)
       return false
--- a/libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
+++ b/libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
@@ -1,7 +1,6 @@
 import {
   deepClone,
   deepFreeze,
-  objKeyMap,
   validateArr,
 } from '@genshin-optimizer/common/util'
 import type {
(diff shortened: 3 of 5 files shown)


libs/common/database/src/lib/DataManagerBase.ts
```diff
@@ -78,7 +78,10 @@ export class DataManagerBase<
       | ((v: StorageValue) => Partial<StorageValue> | void),
     notify = true
   ): boolean {
-    const old = this.getStorage(key)
+    const old =
```
frzyc, line 81: I think this is probably safe enough to be rolledout for all instance of DataManagerBase, instead of creating an one-off here.

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -307,3 +308,33 @@ export function handleArtSetExclusion(
...
+    const y = f(x)
+    if (y) {
+      a.push(y)
+    }
+    reuse &&= x !== undefined && y === x
+  }
+  return reuse ? xs : a
+}
+
+function objectKeyMapInplace<K extends string, T>(
```
frzyc, line 325: Can you add some comments to what the assumption of this function is, and why we should use it over the regular `objKeyMap`? would also be nice to annotate the variables as well

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -307,3 +308,33 @@ export function handleArtSetExclusion(
...
+  ks: readonly K[],
+  o: Record<K, T>,
+  f: (k: K) => T
+) {
+  const oKeys = Object.keys(o)
+  let reuse = oKeys.length == ks.length && ks.every((k, i) => k === oKeys[i])
```
frzyc, line 331: Can shortcut the logic?
```suggestion
  if(oKeys.length == ks.length && ks.every((k, i) => k === oKeys[i])) return o
```

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -307,3 +308,33 @@ export function handleArtSetExclusion(
   }
   return artSetExclusion
 }
+
+function mapFilterInplace<T>(xs: T[], f: (x: T) => T | undefined): T[] {
+  const a: T[] = []
+  let reuse = true
```
frzyc, line 314: Can you predetermine `reuse` with a separate loop? The assumption is that reads are less expensive than push(writes)

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -307,3 +308,33 @@ export function handleArtSetExclusion(
   }
   return artSetExclusion
 }
+
+function mapFilterInplace<T>(xs: T[], f: (x: T) => T | undefined): T[] {
```
frzyc, line 312: Can you add some comments to what the assumption of this function is, and why we should use it over the regular `map.filter`? would also be nice to annotate the variables as well

frzyc: We can also consider moving this function to comon/util function, if we can see more places where we can use this utility

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
frzyc: We can also consider moving this function to comon/util function, if we can see more places where we can use this utility

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
```diff
@@ -193,21 +192,23 @@ export class OptConfigDataManager extends DataManager<
...
       buildDate = 0
     } else {
-      builds = builds
-        .map((build) => {
-          if (typeof build !== 'object') return undefined
-          const { weaponId, artifactIds: artifactIdsRaw } =
-            build as GeneratedBuild
-          if (!this.database.weapons.get(weaponId)) return undefined
-          const artifactIds = objKeyMap(allArtifactSlotKeys, (slotKey) =>
+      builds = mapFilterInplace(builds, (build) => {
```
frzyc, line 195: If I'm understanding your implementation, `builds` won't change if there are no changes from validation, which means that validation is not generating a new `builds` array(that are the same as before). Is that correct?

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
```diff
@@ -73,6 +73,8 @@ type BuildDisplayItemProps = {
   disabled?: boolean
   extraButtonsRight?: JSX.Element
   extraButtonsLeft?: JSX.Element
+  mainStatAssumptionLevel: number
```
frzyc, line 76: So from my understanding, you are prop drilling these 2 props used by `BuildDisplayItem` from optConfig, so that any other changes to optconfig does not trigger rerendering of `BuildDisplayItem`, is that correct?

frzyc (changes requested):
Some minor change requests. Would be nice for the future to get some explanation/rationale on why these changes are made, I can only hazard a guess at why you did some of the things you did.

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
eeeqeee: No, this doesn't validate the object's values

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
eeeqeee: I could, but it would end up calling `f` twice for each array element.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/Components/BuildDisplayItem.tsx
eeeqeee: Yep

libs/common/database/src/lib/DataManagerBase.ts
eeeqeee: iirc it was causing errors when I enabled it for all instances

libs/gi/db/src/Database/DataManagers/OptConfigDataManager.ts
eeeqeee: Yep

frzyc (approved):

Issue #1817, opened 2024-03-22
Standardize comparison when comparing builds


frzyc:
There is a discrepancy between the current/colored comparison build in terms of "build result" vs "tc interface". We should seek to standardize this:
 [current build number] [+/- compare build number]
 
 This should make it so that regardless if you are comparing or not, the "white text" should show the "current build" value

Also, in the codebase, the `DataContext` are referred to as `{data, oldData}`, which is not be best naming, I'd recommend refactoring to `{data, compareData}` to ease this transition.

PR #1828, opened 2024-03-25
Standardize compare number

frzyc:
## Describe your changes
When comparing builds, the "white text" should always be the focused build, and the "delta" colored text will be the diff between the focused/compare build. This prevents the "white text" from changing regardless of the compare state, which should give a more "stable" experience to compare numbers for users.

## Issue or discord link

- Resolves https://github.com/frzyc/genshin-optimizer/issues/1817
## Testing/validation

For screenshot, Left is new, Right is production
Overview screen 
Screenshot 2024-03-25 011002
Theorycraft
Screenshot 2024-03-25 010707
Generated Build
Screenshot 2024-03-25 011418
Compare build (swapping single artifact)
Screenshot 2024-03-25 012342

### DIFF
--- a/apps/frontend/src/app/Components/Character/CharacterEditor/index.tsx
+++ b/apps/frontend/src/app/Components/Character/CharacterEditor/index.tsx
@@ -60,7 +60,7 @@ function CharacterEditorContent({
     return {
       data: charUIData,
       teamData,
-      oldData: undefined,
+      compareData: undefined,
     }
   }, [charUIData, teamData])
 
--- a/apps/frontend/src/app/Components/FieldDisplay.tsx
+++ b/apps/frontend/src/app/Components/FieldDisplay.tsx
@@ -14,7 +14,6 @@ import {
   List,
   ListItem,
   Skeleton,
-  Tooltip,
   Typography,
   styled,
 } from '@mui/material'
@@ -133,15 +132,13 @@ export function NodeFieldDisplay({
       <>
         <span>{valueString(nodeValue, node.info.unit, node.info.fixed)}</span>
         {Math.abs(diff) > 0.0001 && (
-          <Tooltip
+          <BootstrapTooltip
             title={
               <Typography>
                 Compare to{' '}
-                <ColorText color={diff > 0 ? 'success' : 'error'}>
-                  <strong>
-                    {valueString(compareValue, node.info.unit, node.info.fixed)}
-                  </strong>
-                </ColorText>
+                <strong>
+                  {valueString(compareValue, node.info.unit, node.info.fixed)}
+                </strong>
               </Typography>
             }
           >
@@ -166,7 +163,7 @@ export function NodeFieldDisplay({
                 </span>
               )}
             </ColorText>
-          </Tooltip>
+          </BootstrapTooltip>
         )}
       </>
     )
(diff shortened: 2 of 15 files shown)


nguyentvan7:
multi-targets are showing up as red for some reason. Not showing up on beta build 
image

nguyentvan7:
This was already an issue before, but since we are touching this code now:

Should we add a tooltip to the diff value that shows the raw comparison value?
For example, if my current build is 1k dmg
then my compared-to build is 900 dmg

UI will show:
1000 (+100)(+%)

What if we added a tooltip to the `(+100)(+%)` portion that will show `Compared to: 900` or something similar?

To me, without prior knowledge of which "direction" the comparison goes, it is hard to know if the diff value is `current - compare` or `compare - current`

frzyc:
> This was already an issue before, but since we are touching this code now:
> 
> Should we add a tooltip to the diff value that shows the raw comparison value? For example, if my current build is 1k dmg then my compared-to build is 900 dmg
> 
> UI will show: 1000 (+100)(+%)
> 
> What if we added a tooltip to the `(+100)(+%)` portion that will show `Compared to: 900` or something similar?
> 
> To me, without prior knowledge of which "direction" the comparison goes, it is hard to know if the diff value is `current - compare` or `compare - current`

Sure, sounds like we can add it to see how people perceive it, and we can decide what to do next.

apps/frontend/src/app/Components/Character/CharacterEditor/index.tsx
```diff
@@ -60,7 +60,7 @@ function CharacterEditorContent({
     return {
       data: charUIData,
       teamData,
-      oldData: undefined,
```
nguyentvan7, line 63: good change

apps/frontend/src/app/Components/FieldDisplay.tsx
```diff
@@ -121,25 +124,49 @@ export function NodeFieldDisplay({
...
     fieldVal = (
       <>
-        <span>{valueString(oldValue, node.info.unit, node.info.fixed)}</span>
+        <span>{valueString(nodeValue, node.info.unit, node.info.fixed)}</span>
         {Math.abs(diff) > 0.0001 && (
-          <>
-            <ColorText color={diff > 0 ? 'success' : 'error'}>
-              {diff > 0 ? '+' : ''}
-              {valueString(diff, node.info.unit, node.info.fixed)}
+          <Tooltip
```
nguyentvan7, line 136: This tooltip has bad contrast
image
Can we use bootstraptooltip instead for the black background, and to match our other tooltips?

apps/frontend/src/app/Components/FieldDisplay.tsx
```diff
@@ -121,25 +124,49 @@ export function NodeFieldDisplay({
...
+            title={
+              <Typography>
+                Compare to{' '}
+                <ColorText color={diff > 0 ? 'success' : 'error'}>
```
nguyentvan7, line 140: image
This coloring is a little confusing honestly. Might be better without any coloring?

Like it doesn't seem right that the lower number in the tooltip is green for some reason, as if indicating it is better. Im not sure if there would be similar confusion when its flipped

nguyentvan7 (approved):

PR #1842, opened 2024-03-28
reduce lag in opt tab toggles

frzyc:
## Describe your changes

Partially reverts #1819, and update methodology:

- Use [React.memo](https://react.dev/reference/react/memo) to memoize components against needless updates, 
- Add a new hook, `useConstObj` to allow prop checking for updated complex obj data.


apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/index.tsx
```diff
@@ -1155,32 +1141,34 @@ function BuildItemWrapper({
   allowLocationsState: AllowLocationsState
 }) {
   const { t } = useTranslation('page_character_optimize')
-
+  const extraButtonsLeft = useMemo(() => {
```
eeeqeee, line 1144: How much time did this take to render? On which interactions does this save a render?

libs/common/ui/src/hooks/useConstObj.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * Will "retain" the reference to the same object,
+ * as long as the serialized representation remains the same.
+ * @param objInput Input to remain constant
+ * @returns a state reference to the `objInput`,
+ * static as long as the JSON stringlization of `objInput`
+ * remains the same(even if the reference is different).
+ */
+export function useConstObj<T extends object>(objInput: T) {
+  const [obj, setobj] = useState(objInput)
+  const objStr = JSON.stringify(obj)
```
eeeqeee, line 13: https://react.dev/reference/react/Component#shouldcomponentupdate-caveats
> We do not recommend doing deep equality checks or using JSON.stringify in shouldComponentUpdate. It makes performance unpredictable and dependent on the data structure of every prop and state. In the best case, you risk introducing multi-second stalls to your application, and in the worst case you risk crashing it.

eeeqeee (changes requested):
Builds are rerendering when toggling main stats, so this is incomplete

libs/common/ui/src/hooks/useConstObj.tsx
frzyc: yeah, thats in the "should update" logic of class-based components, where there could be unserializable data passed to the component props, like react components and circular references. (which will break JSON.stringify). The object in this case is strongly typed, JSONable(since it come from database), and we are not overriding the existing prop checking within a component, we are just maintaining the `Object.is` reference. I don't think this quote is applicable in this context.

apps/frontend/src/app/PageTeam/CharacterDisplay/Tabs/TabOptimize/index.tsx
frzyc: I believe the `onClick={()=>...}` of the button is triggering renders inside the `BuildDisplayItem`. This memorization prevents that from happening. (an argument can be made to use `useCallback` for that function, but I think this also prevents the `<></>` from triggering updates as well)

frzyc:
> Builds are rerendering when toggling main stats, so this is incomplete

I am not sure how you are determining that, I couldn't see any rerendering of the build list/children when mains are toggled, in the profiler or even directly via console.log in the components.

frzyc:
Addressed those additional child rerenders in `BuildDisplayItem`, by making `optimizationTarget` const.
image

libs/common/ui/src/hooks/useConstObj.tsx
eeeqeee: My concern is deep equality is slower than shallow equality and this cost is paid on every render

libs/common/ui/src/hooks/useConstObj.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * @param objInput Input to remain constant
+ * @returns a state reference to the `objInput`,
+ * static as long as the JSON stringlization of `objInput`
+ * remains the same(even if the reference is different).
+ */
+export function useConstObj<T extends object | undefined>(objInput: T) {
+  const [obj, setobj] = useState(objInput)
+  const objStr = JSON.stringify(obj)
+  useEffect(() => {
+    if (JSON.stringify(objInput) !== objStr) setobj(objInput)
```
eeeqeee, line 15: Nit
```suggestion
  const [obj, setObj] = useState(objInput)
  const objStr = JSON.stringify(obj)
  useEffect(() => {
    if (JSON.stringify(objInput) !== objStr) setObj(objInput)
```

eeeqeee (approved):

frzyc:
> My concern is deep equality is slower than shallow equality and this cost is paid on every render
This is a valid concern, this will only be used in cases like this, where the user can perceive the lag, it won't be used in general UI until it can prove to be an general improvement.

libs/common/ui/src/hooks/useConstObj.tsx
frzyc: Oops, didnt see this till after the merge. Will address it later(TM).

PR #1930, opened 2024-04-09
fix local dev opt build

frzyc:
## Describe your changes

image

### Add a `gi-solver-tc`

Note: the `gi-ui` dependency is due to unit test, and not the actual core code. (Will need to get cleaned up later)
image

## Issue or discord link

- 


nguyentvan7 (approved):
So is the true fix libifying all the teamdata code etc and separating it from all the react hoook stuff?

frzyc:
> So is the true fix libifying all the teamdata code etc and separating it from all the react hoook stuff?

Yes, the fundamental issue is that our WR engine code, which is a dependency for almost every UI-forward lib, has UI dependency, which causes "pure TS" libs to be very difficult to be extracted. I've so far isolated it to `gi-uidata`, but going further would require some delicate refactoring.

PR #1944, opened 2024-04-12
Revive go-next

frzyc:
## Describe your changes

Get go-next to a working state again.
Add `@prisma/nextjs-monorepo-workaround-plugin` to resolve multiple instance of Prisma Client due to hot refresh in NextJS.

## Issue or discord link

- 


libs/gi/ui-next/src/client/ArtifactCard.tsx
```diff
@@ -312,7 +312,7 @@ function Header({
             <ColorText color={mainStatLevel !== level ? 'warning' : undefined}>
               {getMainStatDisplayStr(
                 mainStatKey,
-                rarity as RarityKey,
+                rarity as ArtifactRarity,
```
nguyentvan7, line 315: Why do we have to type assert?

frzyc: Artifact.rarity is a Int (number) from GraphQL interface, `ArtifactRarity` is 3|4|5
https://github.com/frzyc/genshin-optimizer/blob/master/apps/gi-backend/src/app/schema_gen.graphql#L33

nguyentvan7: i thought the benefit of this was it had type safety

nguyentvan7 (approved):

libs/gi/ui-next/src/client/ArtifactCard.tsx
frzyc: getting the type for rarity to be a more strict enum is just extra work that hasnt been done yet, it will still validate it to be a int, which is better than `any`

PR #1958, opened 2024-04-14
Use GO page on go-next

frzyc:
## Describe your changes

Allow using GO's page libs in go-next. This is proof that we no longer need to have dedicated go-next components, and can share components with the main site. 
Currently, Implemented the artifact page. 
image
The main sticking point is the ability to conditionally use NextJS's Image component in NextJS environment.

## Issue or discord link

- 


libs/gi/db/src/thread.ts
```diff
@@ -1 +0,0 @@
-export const defThreads = navigator?.hardwareConcurrency || 4
```
nguyentvan7, line 1: Does this change need to happen here? Max threads being the default feels the best for user perception of the site/how fast it can run

frzyc: The issue is that the `navigator` doesnt exist on server side, so having this API call in the "db" lib, which is suppose to be a pure JS lib, is technically wrong to begin with. I am pretty sure this doesn't change the functionality on the GO side, moving this API call into the UI side.

nguyentvan7: It does change the default number of threads 
image

nguyentvan7: Can you add a `navigator?.hardwareConcurrency || defThreads` to wherever this is used in the regular frontend UI?

nguyentvan7 (commented):
see reply

libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/index.tsx
```diff
@@ -156,15 +156,15 @@ export default function TabBuild() {
...
+  const [{ threads }, setDisplayOptimize] = useState(
     database.displayOptimize.get()
   )
   useEffect(
     () => database.displayOptimize.follow((_r, to) => setDisplayOptimize(to)),
     [database, setDisplayOptimize]
   )
-
-  const maxWorkers = threads > defThreads ? defThreads : threads
+  const nativeThreads = navigator?.hardwareConcurrency || 4
```
frzyc, line 166: @nguyentvan7 
> Can you add a navigator?.hardwareConcurrency || defThreads to wherever this is used in the regular frontend UI?

I did, Its right here.

nguyentvan7: That is just for the dropdown maximum, I was asking for the default selection, here: https://github.com/frzyc/genshin-optimizer/blob/d59b55cbfb785fb2986c0094499c9ea19a8503ad/libs/gi/db/src/Database/DataEntries/DisplayOptimizeEntry.ts#L11

which I guess is also a non-ui library. I still think the max threads should be default, but maybe there's no way to do that

frzyc: I sort of see what you are getting at. I think you have the right idea, we should remove this option from database.

libs/gi/ui/src/hooks/useNumWorkers.tsx
```diff
@@ -0,0 +1,28 @@
+import { clamp } from '@genshin-optimizer/common/util'
+import { useCallback, useEffect, useState } from 'react'
+const LS_KEY = 'num_opt_workers'
+export function useNumWorkers() {
```
frzyc, line 4: Could hoist this into common/ui, so that SRO can use it, however, SRO might have different worker requirements.

nguyentvan7: I can move it later if needed, here is good for now

nguyentvan7 (approved):
Thanks for changing the threads thing

PR #1971, opened 2024-04-16
Rename sr-opt and fix built count being wrong

### LINKED ISSUE #1966: [SRO] # of builds returned exceeds the # of builds proposed to be calculated. Result is also NaN
Image

nguyentvan7:
## Describe your changes

* Renamed library to sr-solver to match go-solver
* Fixed build count going larger than possible because the build list was not being cleared out

## Issue or discord link

- Resolve #1966 

## Testing/validation

Tested optimization, correct # of builds was being generated, results seemed correct.


libs/sr/solver/src/parentWorker.ts
```diff
@@ -92,7 +92,7 @@ async function start({
     relicsBySlot[slot].map(convertRelicToStats)
   )
 
-  const chunkSize = Math.ceil(relicsBySlot.head.length / numWorkers)
```
frzyc, line 95: The worse case for current implementation is that the last worker might not have work, which means every other worker has 1 "work", which means you have less or equal than worker-1 work in total.(I would argue that the number is small enough not to warrent splitting at all, since workers have overhead to spin up. )

For this new implementation, the worse case is the last worker having up to numworker-1 extra work, which can happen at almost any number. 

In terms of a distributed system, this doesnt seem like better behaviour, and imo the correct fix is just allow last worker not have any work, and probably likely implement a min-number on work assignment

nguyentvan7: There are 6 relic slots, I split the work on the first slot only for simplicity.

If the first slot is 28, and there are 8 workers, the old impl will give a `chunkSize` of `ceil(28/8)` = 4.
In this case, first 7 workers will have 4 (7*4=28) 'work', and the last worker will have 0 work.

In the new impl, it will give a chunkSize of `floor(28/8)` = 3.
In this case, first 7 workers will have 3 work, and the last will have (28-21) = 7 work.

In general, I think you are right. I can just make this mildly more complex to more evenly distribute the work, so each the range in work is only +-1 across all workers, as that could be dramatic speed up.

frzyc: Another easy opt could be to split along the relic slots with the highest number of pieces(instead of just head)

nguyentvan7: What is the benefit in this case? less variance in work between workers?

frzyc: not necessarily less variance, just more "fair" splitting ( think extreme cases where there are only 2 head relics for 8 workers)

frzyc (approved):

PR #1983, opened 2024-04-17
Split worker work more evenly

nguyentvan7:
## Describe your changes

Use a slightly more sophisticated method to split worker work more evenly, based on the slot with the most relics.

## Issue or discord link

- 

## Testing/validation

* Verified via logging that work is as evenly distributed as possible for small case (24 total perms) and large case (128 million perms)
* Verified that optimization in both cases reports the correct number of calculated builds

### DIFF
--- a/libs/sr/solver/src/parentWorker.ts
+++ b/libs/sr/solver/src/parentWorker.ts
@@ -92,7 +92,30 @@ async function start({
     relicsBySlot[slot].map(convertRelicToStats)
   )
 
-  const chunkSize = Math.ceil(relicsBySlot.head.length / numWorkers)
+  // Split work on the largest slot to reduce variance in work between workers
+  const { largestSlot } = Object.entries(relicsBySlot).reduce(
+    ({ largestSlot, largestSize }, [currentSlot, relics]) =>
+      relics.length > largestSize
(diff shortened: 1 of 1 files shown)


frzyc (approved):

libs/sr/solver/src/parentWorker.ts
```diff
@@ -92,7 +92,30 @@ async function start({
...
+  )
+  // Split work
+  const chunkedRelicStatsBySlot: Record<RelicSlotKey, RelicStats[]>[] = range(
+    0,
+    numWorkers - 1
+  ).map((worker) =>
+    objKeyMap(allRelicSlotKeys, (slot) =>
+      slot === largestSlot
+        ? // For largest slot, only give ~1/numWorkers amount of relics
+          relicStatsBySlot[slot].filter(
```
frzyc, line 111: What happens in the case where there are more `numWorker` than relics in `largestSlot`? does worker work with an empty slot array?

nguyentvan7: It works as expected. For the 24 permutation case, the largest slot was 3 and I had 8 workers. 8 workers still spun up but only 3 did actual work. Could optimize more by not spinning up those workers. But in this case, the work should be done super fast anyways

PR #1993, opened 2024-04-19
Add SRO conditional display

### LINKED ISSUE #1965: [SRO] Add simple conditional UI based on tags

nguyentvan7:
## Issue or discord link

- Resolve #1965


libs/pando/ui-sheet/src/document.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface FieldsDocument extends BaseDocument {
+  type: 'fields'
+  fields: Field[]
+}
+export interface ConditionalDocument extends BaseDocument {
+  type: 'conditional'
+  conditional: Conditional
+}
+
+export interface BaseDocument {
```
frzyc, line 21: would it make sense for a document to be componsed of multiple of these elements?

for example, if i want a 
```
HEADER
text
fields
conditional
```
all under one document header(like we have for talents)

libs/pando/ui-sheet/src/sheet.ts
```diff
@@ -0,0 +1,5 @@
+import type { Document } from './document'
+
+export type UISheet = {
+  documents: Document[]
```
frzyc, line 4: would it make sense to associate each document with a key of some sort, for easier reverse lookup?

libs/pando/ui-sheet/src/document.ts
nguyentvan7: image
This structure is a 1:1 match/parity of current ui sheet
blue = TextDocument
red = FieldsDocument
yellow = ConditionalDocument

green = optional header

So im not sure what you meant by the HEADER over multiple text/fields/conditional. I believe that is just created within the UI itself, not necessarily in the sheet

frzyc: Oh i misunderstand, I thought a character sheet is one UISheet, but i realized its a map of UIsheets, like
`{ auto:UISheets[], ...}`. I guess the naming convention confused me. Then disregard my comments, I am fine with the format for now.

nguyentvan7: A UI sheet is made up of an array or map of documents. I should adjust the shaping in sheet.ts prob

libs/pando/ui-sheet/src/sheet.ts
nguyentvan7: Unresolving because yea we should map it, similar to current UI sheet structure

PR #2021, opened 2024-04-26
v10 migration patch

frzyc:
## Describe your changes

V10 Migration creates a team+loadout for every character in db.
Amend logic to only create a team+loadout if that character had either mtargets or an opt-target selected. 
This should reduce some of the migration noise/cleanup. 
Teammate loadout for characters with mtargets/optargets are still created, to preserver teammate conditionals.

## Issue or discord link

- 

### DIFF
--- a/libs/gi/db/src/Database/migrate.ts
+++ b/libs/gi/db/src/Database/migrate.ts
@@ -172,6 +172,8 @@ export function migrateGOOD(good: IGOOD & IGO): IGOOD & IGO {
       const optConfig =
         buildSettings.find(({ id }: { id: string }) => id === characterKey) ??
         {}
+      // Only migrate characters with either a mtarget or an opttarget
+      if (!customMultiTarget.length && !optConfig.optimizationTarget) return
       const optConfigId = `optConfig_${optConfigInd++}`
       ;(good as any).optConfigs.push({ ...optConfig, id: optConfigId })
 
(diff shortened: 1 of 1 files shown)


libs/gi/db/src/Database/migrate.ts
```diff
@@ -172,6 +172,8 @@ export function migrateGOOD(good: IGOOD & IGO): IGOOD & IGO {
       const optConfig =
         buildSettings.find(({ id }: { id: string }) => id === characterKey) ??
         {}
+      // Only migrate characters with either a mtarget or an opttarget
+      if (!customMultiTarget?.length && !optConfig.optimizationTarget) return
```
nguyentvan7, line 176: Not sure if we even need the opt target thing. Won't this still end up creating 15 loadouts if a support is in 15 teams and has a opt target?
Or am I reading the logic wrong and this will just create one loadout?

frzyc: merging loadouts is complicated logic that probably can't be safely implemented.

This logic avoids creating teams/loadouts for characters that are in GO, but don't have a opt-target selected. 
If someone has like 60 characters, but only set up optimizer for like 20 characters, it will leave the other 40 characters that are not in any teams of the core 20 characters to be "no loadout/team"

nguyentvan7 (approved):
i'd kinda be down to only migrating a single loadout per char, where the loadout is the one with multi-opt. conditionals and selected opt target are easy to recreate. tho i guess build constraints might want to be migrated idk. either way needs more thought for further progress, but good changes

PR #2025, opened 2024-04-26
Create basic Relics page

### LINKED ISSUE #1995: Move RelicInventory and RelicEditor to Relic page
Should be a new library (see https://github.com/frzyc/genshin-optimizer/wiki/Generating-new-libraries)
Should be placed similarly to GO's Artifacts tab

tiffanynwyeung:
## Describe your changes
- Generate Relics page lib
- Move Relic-related components to specific page
- Standardize styling across Relics and Light Cones pages (in-depth explanation: make it look more like the GO pages by wrapping things in `CardThemed` and removing `Container` components because it's rendering the site UI smaller than usual with `Container` wrapped around it)
- remove `useTranslation` hook being improperly called in sort/filter configs from last PR

## Issue or discord link
Resolves #1995 

## Testing/validation
Relics page (note: this was screenshotted with browser zoomed out b/c currently on smaller screen resolution)
image

Light Cone page
image

### DIFF
--- a/apps/sr-frontend/src/app/Header.tsx
+++ b/apps/sr-frontend/src/app/Header.tsx
@@ -1,7 +1,7 @@
 import { useForceUpdate } from '@genshin-optimizer/common/react-util'
 import { AnvilIcon } from '@genshin-optimizer/common/svgicons'
 import { useDatabaseContext } from '@genshin-optimizer/sr/ui'
-import { Menu as MenuIcon } from '@mui/icons-material'
+import { Diamond, Menu as MenuIcon } from '@mui/icons-material'
 import {
   AppBar,
   Box,
@@ -34,12 +34,35 @@ type ITab = {
   value: string
   textSuffix?: ReactNode
 }
+const relics: ITab = {
+  i18Key: 'tabs.relics',
+  // No SRO icons exist right now, replace later with real relics icon
+  icon: <Diamond />,
+  to: '/relics',
+  value: 'relics',
+  textSuffix: <RelicChip key="relicsAdd" />,
+}
+
 const lightCones: ITab = {
   i18Key: 'tabs.lightcones',
   icon: <AnvilIcon />,
   to: '/lightcones',
   value: 'lightcones',
   textSuffix: <LightConeChip key="lightConeAdd" />,
+}
+
+function RelicChip() {
+  const { database } = useDatabaseContext()
+  const [dirty, setDirty] = useForceUpdate()
+  useEffect(
+    () => database.relics.followAny(() => setDirty()),
+    [database, setDirty]
+  )
+  const total = useMemo(
+    () => dirty && database.relics.keys.length,
+    [dirty, database]
+  )
+  return <Chip label={<strong>{total}</strong>} size="small" />
 }
 
 function LightConeChip() {
@@ -64,7 +87,7 @@ export default function Header({ anchor }: { anchor: string }) {
   )
 }
 
-const maincontent = [lightCones] as const
+const maincontent = [relics, lightCones] as const
 
 function HeaderContent({ anchor }: { anchor: string }) {
   const theme = useTheme()
@@ -144,7 +167,7 @@ function HeaderContent({ anchor }: { anchor: string }) {
   )
 }
 
-const mobileContent = [lightCones] as const
+const mobileContent = [relics, lightCones] as const
 function MobileHeader({
   anchor,
   currentTab,
--- a/apps/sr-frontend/src/app/Header.tsx
+++ b/apps/sr-frontend/src/app/Header.tsx
@@ -36,7 +36,7 @@ type ITab = {
 }
 const relics: ITab = {
   i18Key: 'tabs.relics',
-  // No SRO icons exist right now, replace later with real relics icon
+  // TODO: replace with real relics icon later
   icon: <Diamond />,
   to: '/relics',
   value: 'relics',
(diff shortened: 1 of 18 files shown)


apps/sr-frontend/src/app/Header.tsx
```diff
@@ -42,6 +51,20 @@ const lightCones: ITab = {
   textSuffix: <LightConeChip key="lightConeAdd" />,
 }
 
+function RelicChip() {
+  const { database } = useDatabaseContext()
+  const [dirty, setDirty] = useForceUpdate()
+  useEffect(
+    () => database.relics.followAny(() => setDirty()),
```
frzyc, line 58: This is more of a nit to how we do this in GO, I think these gets rerendered every time we change an artifact/relic, but it should rerender every time an artifact/relic is added/removed (since it only display the total). 

It doesn't have to be this PR, but I think it make sense to use the `DataManagerCallback.reason` string, and check against `'remove' | 'new'`. Would also make sense to store `total` as a state instead of `useForceUpdate`, so when total is set as the same value again, it doesnt trigger a revalue.

libs/sr/page-relics/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "sr-page-relics",
+  "$schema": "../../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "libs/sr/page-relics/src",
+  "projectType": "library",
+  "tags": [],
+  "targets": {
+    "lint": {
```
frzyc, line 8: [request change] `lint` can be removed, since this task can be inferred.

libs/sr/ui/src/LightCone/LightConeEditor/LightConeEditor.tsx
```diff
@@ -124,7 +123,7 @@ export function LightConeEditor({
   }, [lightConeIdToEdit])
 
   return (
-    <Container>
+    <Suspense fallback={false}>
```
frzyc, line 126: Probably a good idea to put something in here, since using `fallback=false` would cause layout shift as element is loaded. I am fine leaving this here for now since we are likely to put this element in a modal later.

frzyc (changes requested):
Some editorial comments, and one request change.

libs/sr/ui/src/Relic/RelicInventory.tsx
```diff
@@ -20,86 +10,89 @@ const numToShowMap = { xs: 10, sm: 12, md: 24, lg: 24, xl: 24 }
...
-  const { relicsIds, totalRelicsNum } = useMemo(() => {
-    const relicsIds = database.relics.keys
-    const totalRelicsNum = relicsIds.length
-    return { relicsIds, totalRelicsNum }
-  }, [database])
+  useEffect(
+    () =>
+      database.relics.followAny(
+        (_k, r) =>
+          ['new', 'remove'.includes(r)] &&
```
frzyc, line 19:
```suggestion
          ['new', 'remove'].includes(r) &&
```

libs/sr/ui/src/Relic/RelicInventory.tsx
```diff
@@ -20,86 +10,89 @@ const numToShowMap = { xs: 10, sm: 12, md: 24, lg: 24, xl: 24 }
...
+          setTotal(database.relics.values.length)
+      ),
+    [database]
+  )
+
+  const relicIds = useMemo(() => {
+    const relics = database.relics.values
+    const relicIds = relics.map((r) => r.id)
+    return relicIds
+  }, [database.relics.values])
```
frzyc, line 29: Using `[database.relics.values]` as the dependency array here would actually trigger a recalc every time, since it would create a new array every time its called. 
```suggestion
  }, [total])
```

libs/sr/util/src/lightcone.ts
```diff
@@ -1,31 +1,32 @@
 import type { FilterConfigs, SortConfigs } from '@genshin-optimizer/common/util'
 import type { ILightCone } from '@genshin-optimizer/sr/srod'
 import { allStats } from '@genshin-optimizer/sr/stats'
-import { useTranslation } from 'react-i18next'
 
 export const lightConeSortKeys = ['level', 'name', 'rarity'] as const
 export type LightConeSortKey = (typeof lightConeSortKeys)[number]
 
+// SRO currently has NO i18n at the moment.
```
frzyc, line 8: add a `// TODO:` here, thats generally how we tag TODOs

frzyc (changes requested):

apps/sr-frontend/src/app/Header.tsx
```diff
@@ -42,17 +50,34 @@ const lightCones: ITab = {
...
+      database.relics.followAny(
+        (_k, r) =>
+          ['new', 'remove'].includes(r) && setTally(database.relics.keys.length)
+      ),
+    [database]
   )
-  const total = useMemo(
-    () => dirty && database.lightCones.keys.length,
-    [database, dirty]
+  const total = useMemo(() => tally, [tally])
```
frzyc, line 64: There is no reason to memoize this, since `tally` is already a state. `useMemo` should be used to "cache" expensive calculations.

apps/sr-frontend/src/app/Header.tsx
```diff
@@ -42,17 +50,34 @@ const lightCones: ITab = {
...
+  useEffect(
+    () =>
+      database.lightCones.followAny(
+        (_k, r) =>
+          ['new', 'remove'].includes(r) &&
+          setTally(database.lightCones.keys.length)
+      ),
+    [database]
   )
+  const total = useMemo(() => tally, [tally])
```
frzyc, line 80: Same here, no need to memoize.

frzyc (changes requested):
Remove the 2 extra `useMemo` and we should be good.

frzyc (approved):
LGTM

PR #2028, opened 2024-04-28
Add Sort and Show/Sort components to common UI library

tiffanynwyeung:
## Describe your changes
- Copy `SortByButton` and `ShowingAndSortOptionSelect` components from GO UI library to common UI library

These two components seem to be safe to copy over to the common UI library, and therefore can be used later by the SRO UI for setting up sorting/filtering UI options. The only thing that might break it is that SRO doesn't have localization atm.

## Issue or discord link
Technically related to https://github.com/frzyc/genshin-optimizer/issues/2030 

## Testing/validation
Nothing calls it at the moment


frzyc (changes requested):
Can you also remove the `SortByButton` and other elements in `gi/ui` lib as well? Also, I am not sure if i18n are added in SRO yet, so these might not be usable for SRO immediately.

frzyc (approved):

Issue #2029, opened 2024-04-28
Error in the Screenshot parsing


LukasHuth:
I started to upload all my artifacts and noticed, that all artifacts with $HP >= 1000$ failed to get parsed. The bug is reproduceable by selecting the "Artifacts" tab, then clicking on "Add New Artifact" and after that pasting an artifact with $HP >= 1000$ (like the images linked below). I also linked a screenshot, of the parsing failing. I suspect a problem with the `,` in the parsing process.

## Example artifact images
![image](https://github.com/frzyc/genshin-optimizer/assets/60547284/69363fda-0161-4be4-88f8-2e69306630df)
![image](https://github.com/frzyc/genshin-optimizer/assets/60547284/c0070b0c-5547-495b-bd2e-f3b60c68c7e1)
![image](https://github.com/frzyc/genshin-optimizer/assets/60547284/b0ad0c6d-ff5a-4f0e-bd3c-4e12d47bc3a1)

## Example of the parsing failing
![image](https://github.com/frzyc/genshin-optimizer/assets/60547284/ac7caeac-84fe-4f12-9b5f-2fbd9a285145)

frzyc:
Interesting. The parsing logic should consider commas. Will need investigation.
https://github.com/frzyc/genshin-optimizer/blob/master/libs/gi/art-scanner/src/lib/parse.ts#L95 (`parseSubstats`)

PR #2052, opened 2024-04-30
Fix the parsing error of Issue #2029

LukasHuth:
## Describe your changes

I changed the replace from , -> . to , -> nothing, because the stat got interpreted as a float and in a later step rounded down to 1
## Issue or discord link
- [Issue 2029 opened by myself](https://github.com/frzyc/genshin-optimizer/issues/2029)

### DIFF
--- a/libs/gi/art-scanner/src/lib/parse.ts
+++ b/libs/gi/art-scanner/src/lib/parse.ts
@@ -108,7 +108,7 @@ export function parseSubstats(texts: string[]): ISubstat[] {
         matches.push({
           key,
           value: parseFloat(
-            match[1].replace(/(,|.)(\d{3})/g, '$2').replace(/,/g, '.').replace(/\.{2,}/g, '.')
+            match[1].replace(/(,|\.)(\d{3})/g, '$2').replace(/,/g, '.').replace(/\.{2,}/g, '.')
           ),
         })
     })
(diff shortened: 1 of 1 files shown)


libs/gi/art-scanner/src/lib/parse.ts
```diff
@@ -108,7 +108,7 @@ export function parseSubstats(texts: string[]): ISubstat[] {
         matches.push({
           key,
           value: parseFloat(
-            match[1].replace(/,/g, '.').replace(/\.{2,}/g, '.')
```
frzyc, line 111: Some European countries use a comma in decimal, like `HP+4,1%`. This code intended to to convert that to `HP+4.1%`.
However, you are right in that the issue exists here, `HP+1,016` -> `HP+1.016`.
Since flat stats are rounded, I think the solution is to only apply the `,`->`''` logic on flat stats (or only on flat HP)

frzyc (changes requested):

libs/gi/art-scanner/src/lib/parse.ts
```diff
@@ -108,7 +108,7 @@ export function parseSubstats(texts: string[]): ISubstat[] {
...
+            match[1].replace(/(,|.)(\d{3})/g, '$2').replace(/,/g, '.').replace(/\.{2,}/g, '.')
```
frzyc, line 111:
```suggestion
            match[1].replace(/(,|\.)(\d{3})/g, '$2').replace(/,/g, '.').replace(/\.{2,}/g, '.')
```

frzyc (changes requested):

frzyc:
Also need to format using `yarn run mini-ci` or `nx format:write`

frzyc (approved):

frzyc:
LGTM, thanks for the fix.

Issue #2054, opened 2024-05-01
Use Equipment confirmation modal for equipping arti/weapons


frzyc:
When swapping artifact/weapons, you are presented with this screen:

<img width="1011" alt="image" src="https://github.com/frzyc/genshin-optimizer/assets/1754901/ea3fb048-8b50-49da-a810-982555a546e5">
Clicking on the green will pop up a comparison, and then upon equip, will show this screen:
<img width="1009" alt="image" src="https://github.com/frzyc/genshin-optimizer/assets/1754901/8cb0dc9a-20c1-41d1-a83e-c10a7d45e1e9">
whereas clicking on the Red will swap without popping up the build confirmation screen.

Since this build confirmation screen is very useful, we should standardize it, and apply it to both equipment path.

Question: does this make sense to also apply this to the equipment dropdown in artifactcard/weaponcard as well?

frzyc:
Here is another thought, can be combine the "Confirm Equipment" Modal with the "build diff" modal? they share a lot of the same UI elements, it is not a stretch to combine them. Would require breaking up `BuildEquipModal` into its pieces so its easier to be put together for different context.

arquelion:
Probably a good idea to standardize the build swap confirmation for consistency's sake as well. I dunno how often users accidentally equip an artifact/weapon or equip to the wrong build whether by directly clicking on it or misclicking in the dropdown, but in theory, this would guard against that too. That being said, `BuildDisplayItem` didn't have the option to equip at all prior to teams/loadouts, and the comparison UI piggybacks off it, so maybe it's worth asking whether having the option to equip both through clicking on the card itself and clicking "Compare to current build" then "Equip to current build" is a little redundant now.

I can see some potential value in merging `BuildEquipModal` in with the confirmation since presumably the reason a user might want to see an individual comparison is that they're already considering an artifact/weapon swap (but doesn't necessarily care about being optimal, else they would be in the Optimize tab and not here), and if they're satisfied then they can immediately swap without clicking on another button first. However, like I mentioned above, the comparison UI currently piggybacks off `BuildDisplayItem`, so if going this route, I'd probably pull the comparison itself out into its own thing and create a new modal that combines the comparison with elements from `BuildEquipModal` instead of actually directly inserting the pieces into `BuildDisplayItem` and then having to conditionally render all of those to prevent them from showing up in the builds under Optimize. I do worry if it'll bloat the UI too much though...

PR #2055, opened 2024-05-01
Equip filter is custom by default, add alert for teammate equipment in builds

### LINKED ISSUE #1625: Artifact Equipment filter update
In the old character-centric implementation of GO, the "equipped" build is used to resolve equipment of artifacts across several character. It made sense when optimizing, to use the non-equipped artifacts by default. Users can then opt-in to allow taking of artifacts from specific characters.

In the team loadout space, the usage of this feature is somewhat different. There is already a feature that by default disallow taking artifacts in teammates' active builds, which was partially what the equipment filter was trying to address(resolution of artifact equipment within a team). Since users can also save builds into loadouts for future reference, the equipment filter feature should be readdressed.

1. ~~adding a config button in artifact detail in generated build, next to single character exclusions, that shows the equipment filter (including tally). This should allow users to exclude singular equipment location, while being able to edit the whole list.~~ (probably overkill)
2. Remove the all/custom/none state of the artifact equipment filter. This should make it so that it is "all" by default(no one excluded). The none/all states are less relevant, and users should be encouraged to generate and save builds regardless of equipment.
3. when the "use teamate's active build" toggle is enabled, if a generated build has an artifact on an active build in the team, should an warning, similiar to https://github.com/frzyc/genshin-optimizer/pull/1556

frzyc:
## Describe your changes

- Make equipment filter "custom" by default, and remove other options (simplify logic)
- Add alert if build has artifact equipped on teammates

## Issue or discord link

- Resolves #1625

## Testing/validation
image
image

Screenshot 2024-05-01 145855


nguyentvan7 (commented):
Screenshot_20240501_210951_Chrome
Screenshot_20240501_210700_Chrome

I can't see the indicators for some reason
[Van_2024-04-19_02-59-01.json](https://github.com/frzyc/genshin-optimizer/files/15183668/Van_2024-04-19_02-59-01.json)

Using ayakas only team, on the current settings

I can see the rest of the changes tho

frzyc:
@nguyentvan7 I am not sure what the issue you have is. In the build you selected, you have an artifact equipped by ganyu, but ganyu is not part of your akaya team, so there is no "teammate equipped" indicator? If optimizer generated an artifact equipped by your kazuha, shenhe or layla, then the indicator will show.

nguyentvan7:
I see it's for teammates only. If allowing from teammates is opt in only, why even have the alert specifically for that case, and not the case that is default enabled? I feel like it makes more sense to alert the user for default enabled cases, in case they missed it

nguyentvan7:
Could you add the teammate equipped alert to the build swap preview too please?
Screenshot_20240501_210700_Chrome

frzyc:
> I see it's for teammates only. If allowing from teammates is opt in only, why even have the alert specifically for that case, and not the case that is default enabled? I feel like it makes more sense to alert the user for default enabled cases, in case they missed it

Teammates make sense, because at the end of the day, you want to build a team with a valid, non-conflict build for each teammate, this warning helps resolve the build inside the team.

Creating and saving builds using equipped artifacts from outside the team is a behavior that we want to encourage, hence the "all" filter by default, showing a warning seems to go against that.

> Could you add the teammate equipped alert to the build swap preview too please?
I anticipate a larger overhaul of this UI https://github.com/frzyc/genshin-optimizer/issues/2054, and I would like to address this UI in that context, and not in this PR.

nguyentvan7 (approved):
I don't really get the reasoning but code is good enough

PR #2064, opened 2024-05-03
Add team buffs and bonus stats to stats display

frzyc:
## Issue or discord link

- 

## Testing/validation

image

### DIFF
--- a/libs/gi/page-team/src/CharacterDisplay/FormulaModal.tsx
+++ b/libs/gi/page-team/src/CharacterDisplay/FormulaModal.tsx
@@ -1,7 +1,6 @@
 import {
   CardHeaderCustom,
   CardThemed,
-  ImgIcon,
   ModalWrapper,
   SqBadge,
 } from '@genshin-optimizer/common/ui'
@@ -105,7 +104,7 @@ function FormulaCalc({
   return (
     <CardThemed bgt="light" sx={{ mb: 1 }}>
       <CardHeaderCustom
-        avatar={icon && <ImgIcon size={2} src={icon} />}
+        avatar={icon}
         title={title}
         action={action && <SqBadge>{action}</SqBadge>}
       />
--- a/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/Components/OptimizationTargetSelector.tsx
+++ b/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/Components/OptimizationTargetSelector.tsx
@@ -1,5 +1,5 @@
 import { useBoolState } from '@genshin-optimizer/common/react-util'
-import { ImgIcon, SqBadge } from '@genshin-optimizer/common/ui'
+import { SqBadge } from '@genshin-optimizer/common/ui'
 import { objPathValue } from '@genshin-optimizer/common/util'
 import { useDatabase } from '@genshin-optimizer/gi/db-ui'
 import {
@@ -70,7 +70,7 @@ export default function OptimizationTargetSelector({
   } = (node?.info && resolveInfo(node?.info)) ?? {}
 
   const suffixDisplay = textSuffix && <span> {textSuffix}</span>
-  const iconDisplay = icon ? <ImgIcon src={icon} size={2} /> : nodeIcon
+  const iconDisplay = icon ? icon : nodeIcon
   return (
     <>
       <Button
--- a/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/Components/TargetSelectorModal.tsx
+++ b/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/Components/TargetSelectorModal.tsx
@@ -1,9 +1,4 @@
-import {
-  CardThemed,
-  ImgIcon,
-  ModalWrapper,
-  SqBadge,
-} from '@genshin-optimizer/common/ui'
+import { CardThemed, ModalWrapper, SqBadge } from '@genshin-optimizer/common/ui'
 import { useDatabase } from '@genshin-optimizer/gi/db-ui'
 import {
   DataContext,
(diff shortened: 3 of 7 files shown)


nguyentvan7:
I don't really like the idea of this without some distinction between stats and actual talent values

Or without the ability to hide specific things/collapse them someway? It's a ton of clutter imo

frzyc:
> I don't really like the idea of this without some distinction between stats and actual talent values
> 
> Or without the ability to hide specific things/collapse them someway? It's a ton of clutter imo

I agree on the clutter part, I have some ideas on reducing the clutter, since we are likely to increase the amount of content we show in the stat display in the future.

I guess i sort of get what you meant by "distinction between stats and actual talent values", I am just not sure what would be the way to go. We've always had basic stats as a section, so adding 2 more is not too crazy of a stretch.

nguyentvan7:
I think we should do those clutter reduction steps before doing this

Distinction between stats and talent values was something you had discussed with Kite and agreed with iirc. I forgot what their mockup was for that tho

frzyc:
> I think we should do those clutter reduction steps before doing this
> 
> Distinction between stats and talent values was something you had discussed with Kite and agreed with iirc. I forgot what their mockup was for that tho

I understand where you are coming from, I don't agree with the suggested course of action. I don't have a concrete plan with decluttering, and I won't really know if decluttering is the right solution to this concern. I think getting feedback from the general user base will influence which decision to mitigate. My contemplative solutions towards decluttering will be somewhat sizable changes and I need to be convinced it's worth the effort.

nguyentvan7:
I still don't really get the purpose of adding all these stats here. They are visible elsewhere, which is why I don't see the need to bulk up this display so much

nguyentvan7:
Screenshot_20240503_081943_Chrome
Screenshot_20240503_081921_Chrome

Ignoring all the 0s being shown for some reason, this is just a ton of wasted space once you add something like Zhongli or a 4 element team

frzyc:
I want to have a centralized display for all the stats and numbers. Bonus stats and team buffs are indeed shown in other parts of the UI, but they are generally somewhat hidden(team setting, uncollapse for each char; loadout setting, edit bonus stats). 

All the 0s are unintended, I will address removing them.

nguyentvan7:
But you can see those stats in the formula already. By this logic we should be showing distinction between self conditional stats too or even base stat vs arti/wep stats

frzyc:
Can you explain to me what you think should be part of this display? I think we have a mismatch in terms of what we individually think is the purpose and intention of this piece of UI?

nguyentvan7:
I think what it has now is appropriate. Final stats and outcomes/DMG/heals/etc for the current character. Showing teambuffs and bonus stats is not a direct final outcome of the character, and can be viewed elsewhere

frzyc:
Well, I don't think we've explicitly declared that the stats here are "final outcomes based", I think that statement stands in general, but we do display some intermediary numbers as well, like shenhe's normal attack bonus(which shows up for teammates). If we want to draw the line, then those non-outcome numbers should show in talent tab, but not the "final stat display". I don't think we have the granularity in our system to fully support this, nor do I think it's worth the effort to do so. 

I also don't like the argument that because the bonus stats/team buffs are shown in the formula of targets, it shouldn't be shown here. I don't think people will necessarily want to infer what their bonus stats/teambuffs are from the relevant formulas. Showing them here will help users understand at least from the input side what can contribute to their final results, instead of tracing backwards from their final results. 

This also allows for more dynamic expression here as well, since users can compare changes to the team buffs due to different(artifact/weapon) builds. If you want to pursue having comparisons across different loadouts, this will be necessary as well.

nguyentvan7:
Shenhe DMG bonus is an outcome of shenhe and only shows up on shenhes overview tab

There are numerous other things that contribute to a final DMG stat that still won't be covered by this (artifact sets, weapon stats, self conditionals, talent level, talent mechanics)

The only team buff that changes based on someone elses build is Nahida A1. Otherwise, any other scaling teambuffs will show automatically on the character overview page for the character it scales from (e.g. shenhe teambuff shows on her own overview page)

frzyc:
> There are numerous other things that contribute to a final DMG stat that still won't be covered by this (artifact sets, weapon stats, self conditionals, talent level, talent mechanics)

Artifact set, weapon stats are shown in the build above the stat display, self-conditionals and other conditionals are probably something we can tackle adding once we migrate to pando, since there will be more standardization, character stats like level/talents are something that I want to address as well(I actually have a good idea for that). All these are good points, but I won't be able to add them all in this one PR. I think once these are added in, the stat display will be a very "screenshottable" component, that has all the information.

frzyc:
Hmm, this is weird, locally the team buffs are fine, but on PR deployment it has all the 0 fields
image

nguyentvan7:
image

I still don't really like this because now the distinction between stat and targets is even more blurry. before, stats were always top-left. now they will be mixed in, and we had agreed that stats being mixed here (even in a consistent spot) was confusing for the user. Now that it is mixed in, its even more confusing

frzyc:
> image
> 
> I still don't really like this because now the distinction between stat and targets is even more blurry. before, stats were always top-left. now they will be mixed in, and we had agreed that stats being mixed here (even in a consistent spot) was confusing for the user. Now that it is mixed in, its even more confusing

I see. we could make it so that we dedicate the left column just for stats (base, bonus, teambuffs), what about stat displays outside of the 3 just mentioned? like these:
image
image

nguyentvan7:
I guess I still considered those different because they are affected by hutaos final stats. but you're right, those could still be confusing to new users

frzyc:
> I guess I still considered those different because they are affected by hutaos final stats. but you're right, those could still be confusing to new users

Eitherway, I understand that from your perspective people may confuse these additions (stats fields) to be confusing, since they are not "outputs". I can understand where you are coming from. My perspective is that these fields were never intended to be only "outputs", but rather a place to put relevant calculated numbers. (not sure if you care about the perspective of the creator of this UI). I am not really agreeing with your enforcement of your perspective(that everything here should be "outcome" numbers), since that was not my intention with this UI in the first place. 

I am still for releasing the UI additions in this PR as is, and get feedback on how users perceive their addition. I will pay extra attention to seeing whether people are confused about the type of the number, since its your main concern.

nguyentvan7 (approved):

PR #2108, opened 2024-05-12
Granular team exports

frzyc:
## Describe your changes

Add more granular options to exporting team, So the exporter can choose what they want to be exported.
- selectable equipped build
- selectable builds
- selectable tc builds
- selectable mtargets

Special handling for selected mtarget as opt target

## Issue or discord link

- 

## Testing/validation

Should handling exporting a mtarget, and the imported team still links to that mtarget

Screenshot 2024-05-14 054613

### DIFF
--- a/libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
+++ b/libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
@@ -96,10 +96,10 @@ export default function TeamExportModal({
       .then(() => alert('Copied team data to clipboard.'))
       .catch(console.error)
   }
-  const [selEverything, setSelEverything] = useState(true)
+  const [selAll, setSelAll] = useState(true)
 
-  const unselectEverything = () => {
-    setSelEverything(true)
+  const onUnselAll = () => {
+    setSelAll(true)
     setLoadoutDataExportSetting(
       loadoutData.map((loadoutDatum) =>
         loadoutDatum
@@ -109,8 +109,8 @@ export default function TeamExportModal({
     )
   }
 
-  const selectEverything = () => {
-    setSelEverything(false)
+  const onSelAll = () => {
+    setSelAll(false)
     setLoadoutDataExportSetting(
       loadoutData.map((loadoutDatum) => {
         if (!loadoutDatum) return defLoadoutExportSetting()
@@ -155,11 +155,8 @@ export default function TeamExportModal({
               and loadout data (bonus stats, enemy config, optimize config) are
               exported. All exported non-TC builds are converted to TC builds.
             </Alert>
-            <Button
-              color="info"
-              onClick={selEverything ? selectEverything : unselectEverything}
-            >
-              {selEverything ? 'Select Everything' : 'Unselect Everything'}
+            <Button color="info" onClick={selAll ? onSelAll : onUnselAll}>
+              {selAll ? 'Select All' : 'Unselect All'}
             </Button>
           </Box>
 
--- a/libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
+++ b/libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
@@ -18,6 +18,8 @@ import {
   CustomMultiTargetIcon,
   FieldDisplayList,
 } from '@genshin-optimizer/gi/ui'
+import CheckBoxIcon from '@mui/icons-material/CheckBox'
+import CheckBoxOutlineBlankIcon from '@mui/icons-material/CheckBoxOutlineBlank'
 import CloseIcon from '@mui/icons-material/Close'
 import InfoIcon from '@mui/icons-material/Info'
 import {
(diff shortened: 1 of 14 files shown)


nguyentvan7:
Any thoughts on a quick button on the loadout header to select all for the loadout?
And another quick button to just export everything on the page?

frzyc:
> Any thoughts on a quick button on the loadout header to select all for the loadout? And another quick button to just export everything on the page?

I'm not opposed to a "select everything" button, lets see how hard it is to add it in.

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+  onHide,
+  teamId,
+}: {
+  show: boolean
+  onHide: () => void
+  teamId: string
+}) {
+  const database = useDatabase()
+  const team = useTeam(teamId)!
+  const { loadoutData } = team
```
nguyentvan7, line 52: doesnt have to be this PR, but selecting multiple loadouts would be cool too. For example, sharing a Hutao + XQ team, where Hutao has 2 loadouts, one for shimenawa optimization, and one for non-shimenawa optimization (different opt target, set config, set filters, multi-opt, possibly ER constraints).

Similarly for XQ, there could be a Fav vs Sac loadout, which will have different multi-opt, and different ER constraint

nguyentvan7 (changes requested):

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
frzyc: Seems kind of complicated. since this is on the team setting page, I think exiting the modal and selecting the team would be the way to go.

nguyentvan7: but for this case, you go from a single export/import in your guide to 4 separate team exports/imports

shim + sac
shim + fav
no shim + sac
no shim + fav

singular import + swapping loadouts sounds simpler for a user than 2 import + team swap

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+                  </Grid>
+                )
+            )}
+          </Grid>
+        </CardContent>
+        <Divider />
+        <CardContent
+          sx={{ display: 'flex', gap: 2, justifyContent: 'flex-end' }}
+        >
+          <Button
```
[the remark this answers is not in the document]
frzyc: I tried with the check mark, but the check mark/status doesnt actually update as the data becomes partially selected(and detecting it would be more complicated than i care to make it). I can try moving the button up.

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
nguyentvan7: discussed in discord, should keep them as separate team exports for now, but is something we can think about if it becomes more common

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
```diff
@@ -0,0 +1,333 @@
...
+              exported. All exported non-TC builds are converted to TC builds.
+            </Alert>
+            <Button
+              color="info"
+              onClick={selEverything ? selectEverything : unselectEverything}
+              startIcon={
+                selEverything ? <CheckBoxIcon /> : <CheckBoxOutlineBlankIcon />
+              }
+            >
+              {selEverything ? 'Select Everything' : 'Unselect Everything'}
```
nguyentvan7, line 131:
```suggestion
              'Select All'
```
Button text shouldn't change, since the icon is changing

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
```diff
@@ -0,0 +1,333 @@
...
+            <Alert severity="info">
+              Export the team data to be imported by another user. All the team
+              and loadout data (bonus stats, enemy config, optimize config) are
+              exported. All exported non-TC builds are converted to TC builds.
+            </Alert>
+            <Button
+              color="info"
+              onClick={selEverything ? selectEverything : unselectEverything}
+              startIcon={
+                selEverything ? <CheckBoxIcon /> : <CheckBoxOutlineBlankIcon />
```
nguyentvan7, line 128:
```suggestion
                selEverything ? <CheckBoxOutlineBlankIcon /> : <CheckBoxIcon/>
```

The icon should represent the current state

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
```diff
@@ -0,0 +1,333 @@
...
+              gap: 1,
+              flexDirection: isMobile ? 'column' : 'row',
+            }}
+          >
+            <Alert severity="info">
+              Export the team data to be imported by another user. All the team
+              and loadout data (bonus stats, enemy config, optimize config) are
+              exported. All exported non-TC builds are converted to TC builds.
+            </Alert>
+            <Button
```
nguyentvan7, line 124: It is possibly confusing if the user manually selects everything, since like you said, the button state/checkbox tracking isn't completely linked together. But good enough

nguyentvan7 (changes requested):

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
frzyc: Actually i prefer the icon represent the "Action state", since its not possible to represent the "current state", due to partial selection.

frzyc:
Additional cornercase: if a mtarget is selected as opt target, need to enforce selecting it.

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
nguyentvan7: I found this very confusing because this breaks the consistency in the rest of our UI where current state is reflected by the image. There have also been a few times we reworded buttons to facilitate keeping this consistency

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
frzyc: yeah user would have to click it twice, preventing this behaviour would be spaghetti

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
frzyc: sounds like removing the icon is the way to go then.

libs/gi/page-team/src/TeamSetting/TeamExportModal.tsx
```diff
@@ -0,0 +1,417 @@
+import {
```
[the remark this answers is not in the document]
frzyc: Yeah, let's raise an issue and address it later.

nguyentvan7 (approved):

PR #2122, opened 2024-05-14
Update Team removal modal to selectable loadouts

frzyc:
## Describe your changes

I noticed that people would delete the team caused by the v10 migration(but the loadouts would still remain).
and if people adopt importing of teams, they would have 4 new loadouts from importing that team. deleting this "Team" would mean deleting the 4 loadouts separately, then the team.

Add a modal so user can select loadouts to delete when deleting a team.

Depends on changes from #2108 

## Issue or discord link

- 

## Testing/validation

Screenshot 2024-05-14 132939

### DIFF
--- a/libs/gi/page-team/src/TeamSetting/TeamDelModal.tsx
+++ b/libs/gi/page-team/src/TeamSetting/TeamDelModal.tsx
@@ -1,0 +1,175 @@
+import {
+  BootstrapTooltip,
+  CardThemed,
+  ColorText,
+  ModalWrapper,
+} from '@genshin-optimizer/common/ui'
+import { notEmpty, toggleArr } from '@genshin-optimizer/common/util'
+import type { Team } from '@genshin-optimizer/gi/db'
+import { useDatabase, useTeam, useTeamChar } from '@genshin-optimizer/gi/db-ui'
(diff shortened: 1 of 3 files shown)


libs/gi/page-team/src/TeamSetting/TeamDelModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+        bgt="light"
+        sx={{ border: selected ? '2px red solid' : undefined }}
+      >
+        <CardActionArea onClick={onClick}>
+          <LoadoutHeaderContent teamCharId={teamCharId}>
+            <ColorText color={inTeams.length === 1 ? 'success' : 'warning'}>
+              <Typography>
+                {inTeams.length === 1
+                  ? 'Only in current team'
+                  : `In ${inTeams.length} teams`}
```
nguyentvan7, line 167: Tooltip for team names? Or too much clutter?

nguyentvan7 (approved):

libs/gi/page-team/src/TeamSetting/TeamDelModal.tsx
frzyc: probably, I think the value here is to indicate if loadouts are in more than one team or not, so not 100% sure about showing all of them yet.

Issue #2039, opened 2024-04-30
Use TextField for artifact substat input


frzyc:
<img width="481" alt="image" src="https://github.com/frzyc/genshin-optimizer/assets/1754901/9d67e6c8-9062-4d7f-a00c-62758cce55ed">

The "colored input" is sort of out dated. Probably need to redesign this:
- use [TextField](https://mui.com/material-ui/react-text-field/) instead of the colored "button input"
- Fix the width somehow so the text fields line up vertically
- Maybe use [TextFieldLazy](https://github.com/frzyc/genshin-optimizer/blob/master/libs/common/ui/src/components/TextFieldLazy.tsx)?

chengxi600:
I can take a look into this.

chengxi600:
ah ok, ill come back to this after NumberInputLazy is merged

PR #2146, opened 2024-05-18
NumberInputLazy for artifact substat input

chengxi600:
## Describe your changes

Replaced CustomNumberInput with NumberInputLazy

## Issue or discord link

Resolves https://github.com/frzyc/genshin-optimizer/issues/2039
## Testing/validation

https://github.com/frzyc/genshin-optimizer/assets/23622319/f671a1bc-bc72-4450-adfe-5752c66e9acd

### DIFF
--- a/libs/gi/ui/src/components/artifact/editor/SubstatInput.tsx
+++ b/libs/gi/ui/src/components/artifact/editor/SubstatInput.tsx
@@ -90,7 +90,7 @@ export function SubstatInput({
 
   return (
     <CardThemed bgt="light">
-      <Box sx={{ display: 'flex' }}>
+      <Box sx={{ display: 'flex', height: '5vh' }}>
         <ButtonGroup size="small" sx={{ width: '100%', display: 'flex' }}>
           <DropdownButton
             startIcon={key ? <StatIcon statKey={key} /> : undefined}
@@ -128,7 +128,7 @@ export function SubstatInput({
                 </MenuItem>
               ))}
           </DropdownButton>
-          <Box
+          <NumberInputLazy
             sx={{
               flexBasis: 30,
               flexGrow: 1,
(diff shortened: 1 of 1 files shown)


libs/gi/ui/src/components/artifact/editor/SubstatInput.tsx
```diff
@@ -129,26 +128,31 @@ export function SubstatInput({
                 </MenuItem>
               ))}
           </DropdownButton>
-          <CustomNumberInputButtonGroupWrapper
-            sx={{ flexBasis: 30, flexGrow: 1 }}
+          <Box
```
frzyc, line 131: Can you try putting the `NumberInputLazy` here without the `Box` wrapper? You can apply `sx` directly to the `NumberInputLazy`

libs/gi/ui/src/components/artifact/editor/SubstatInput.tsx
```diff
@@ -129,26 +128,31 @@ export function SubstatInput({
...
                 setSubstat(index, { key, value: value ?? 0 })
               }
               disabled={!key}
               error={!!error}
-              sx={{
-                px: 1,
-              }}
               inputProps={{
-                sx: { textAlign: 'right' },
+                sx: { textAlign: 'right', height: '1vh' },
```
frzyc, line 152: What does the `1vh` here do?

frzyc (changes requested):

libs/gi/ui/src/components/artifact/editor/SubstatInput.tsx
chengxi600: was trying to set the size to be what it was before, moved it to the parent Box

frzyc (changes requested):
I assumed this is caused by the `5vh` in height for the parent wrapper
image
I am not sure why you needed to set the height here manually?

chengxi600:
> I assumed this is caused by the `5vh` in height for the parent wrapper image I am not sure why you needed to set the height here manually?

without setting the height it looks like this for me:
Screenshot 2024-05-18 at 12 09 30 PM

with 5vh:
Screenshot 2024-05-18 at 12 10 02 PM

would it be better to just use fixed height or just not set any height?

frzyc:
I think `vh` is fundamentally the wrong unit here, because its a relative unit to the viewheight, which can be different across different screensizes.
The height of the element changes as the view height is changed(by using console)

Screenshot 2024-05-18 212326

image

frzyc (approved):
LGTM

PR #2184, opened 2024-05-31
Custom Target Expressions

aurceive:
## Describe your changes

I added support for custom target expressions and a convenient editor.
image

Since the last PR I have found at least one indisputably correct way to use CTE.

Example: https://discord.com/channels/785153694478893126/1035719815622561892/1244630283052126298
Another example: https://discord.com/channels/785153694478893126/1035720633167921313/1237378233666633850

## Testing/validation

The optimizer sometimes refuses to work with some large expressions. Like this:
image

This can be bypassed by slightly changing the form of the expression without changing the essence.
image

Also, this case is calculated incorrectly. Although when it is used as part of a larger expression, it is calculated as expected.
image

I can't fix it. It's an engine issue. I preferred to forget about it until the transition to Pando.


frzyc:
@tisseandorius I am still not convinced that Custom Target expression is the solution towards somewhat niche usecases. Can you elaborate on which usecases that existing multi-target features cannot cover? I am not sure I can understand from the discord link you sent.

tisseandorius:
I don't know why, but since this PR was published I have never encountered the optimizer refusing to work when I used CTE.
I also checked the example of the incorrect calculation, now it returns the expected result.

PR #2261, opened 2024-06-23
update header and theme

frzyc:
## Issue or discord link

- 

## Testing/validation

New(left) vs. Old(right)
Screenshot 2024-06-23 001905

Mobile
image


apps/frontend/src/app/Header.tsx
```diff
@@ -217,37 +220,34 @@ function HeaderContent({ anchor }: { anchor: string }) {
               transition: 'background-color 0.5s ease',
               backgroundColor: 'rgba(255,255,255,0.1)',
             },
-          }}
+            '& .Mui-selected': {
```
nguyentvan7, line 223: I think the text is supposed to be bolded as well
design doc:
image

site:
image

frzyc: I did try to do that, but bolding the text makes it slightly wider, so it causes a tiny bit of layout shift

nguyentvan7: Can you try with pseudoelement with prebolded text? I think the current design is a bit hard to notice which tab you're on even with the colorization

https://stackoverflow.com/questions/556153/inline-elements-shifting-when-made-bold-on-hover

frzyc: I've tried that, and many other solutions like fake bolding with dropshadow and manual letter-spacing. I've not gotten it to work and I don't want to spend more time on it.
also, you can't seriously argue that you can't see which tab is activated here if the text isn't bolded?

image

nguyentvan7 (approved):
/shrug

PR #2301, opened 2024-07-07
add thumbnail images to archive command embeds

Godlander:
## Describe your changes

add thumbnail images to archive command embeds

## Testing/validation

image

### DIFF
--- a/apps/somnia/src/assets/assets.ts
+++ b/apps/somnia/src/assets/assets.ts
@@ -1,35 +1,11 @@
+// TODO: move to combined lib with frontend assets
 export const rarityColors = [0x818486, 0x5a977a, 0x5987ad, 0x9470bb, 0xc87c24]
-export const elementAssets = {
-  none: {
-    img: 'https://api.ambr.top/assets/UI/UI_Icon_Item_Temp.png',
-    color: 0xaaaaaa,
-  },
-  anemo: {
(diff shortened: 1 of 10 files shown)


apps/somnia/src/assets/assets.ts
```diff
@@ -0,0 +1,35 @@
+export const rarityColors = [0x818486, 0x5a977a, 0x5987ad, 0x9470bb, 0xc87c24]
+export const elementAssets = {
+  none: {
+    img: 'https://api.ambr.top/assets/UI/UI_Icon_Item_Temp.png',
```
nguyentvan7, line 4: should also add these to AssetData.ts

apps/somnia/src/assets/assets.ts
```diff
@@ -0,0 +1,35 @@
+export const rarityColors = [0x818486, 0x5a977a, 0x5987ad, 0x9470bb, 0xc87c24]
```
nguyentvan7, line 1: we have these in libs/gi/theme/src/Theme.tsx

If we want to be consistent, we can have a libs/gi/color library or something so frontend and bot can share the same rarity colors

nguyentvan7 (changes requested):
various style comments

apps/somnia/src/assets/assets.ts
Godlander: i think that would be better, bot cant use tsx

PR #2293, opened 2024-07-01
Fixed issue #2241 GO team page title shows tabs.undefined

0sani:
## Describe your changes

Changed default behavior when `tab` is null or `characterKey` is null. 

When `characterKey` is null (when on team settings page), formatted as "\<Team Name\> - Team Settings - Genshin Optimizer". Note that `tab` will always be null in this case. 

When `characterKey` is not null and `tab` is null, this means that we're on the "Loadout/Build" tab, so formatted appropriately.

Do note that there's another issue with the Artifact Upgrader tab, where it doesn't display properly, but I'm not familiar enough with the codebase to fix this. 

## Issue or discord link

- Fix <https://github.com/frzyc/genshin-optimizer/issues/2241>

## Testing/validation

Two of the fixed pages, and the one that's still broken, I can make another issue for that one.

image

image

image

## Checklist before requesting a review (leave this PR as draft if any part of this list is not done.)

- [x] I have commented my code in hard-to understand areas.
- [x] I have made corresponding changes to README or wiki.
- [x] For front-end changes, I have updated the corresponding English translations.
- [x] I have run `yarn run mini-ci` locally to validate format and lint.
- [x] If I have added a new library or app, I have updated the deployment scripts to ignore changes as needed

PR #2254, opened 2024-06-20
Updata `teamName` & `teamDesc` on `team` change

HassnHamada:
## Issue or discord link
Resolve: #2251

## Testing/validation
https://github.com/frzyc/genshin-optimizer/assets/44922829/a9369e92-acc3-4971-a27b-6ba313bf9074

## Checklist before requesting a review (leave this PR as draft if any part of this list is not done.)

- [ ] I have commented my code in hard-to understand areas.
- [ ] I have made corresponding changes to README or wiki.
- [ ] For front-end changes, I have updated the corresponding English translations.
- [ ] I have run `yarn run mini-ci` locally to validate format and lint.
- [ ] If I have added a new library or app, I have updated the deployment scripts to ignore changes as needed

PR #2306, opened 2024-07-10
[SRO] Add team page text changes from GO

tiffanynwyeung:
## Describe your changes

- Adds text changes to SRO from GO, specifically:
  * fix team name/desc to show correct names if team changes (#2254)
  * fix team page document title with correct translations (#2293)
- Renames `header` i18n translation call to `ui` in `sr-frontend`: right now, GO uses `ui` for the same translated material, and the `useTitle` hook in `libs/common` relies on it as a result. With it living in `header` for SRO, `useTitle` doesn't know where to call it, so document titles all break if not in `ui`.
  * Alternative solution: break it apart such that `pageTitle` lives exclusively in `ui`. However, the only relevant translation for `header` is the tabs, which is pretty small. Not sure if leaving it scoped exclusively to that degree would be a good idea?


apps/sr-frontend/src/app/Header.tsx
```diff
@@ -148,7 +148,7 @@ function HeaderContent({ anchor }: { anchor: string }) {
   const theme = useTheme()
   const isXL = useMediaQuery(theme.breakpoints.up('xl'))
   const isMobile = useMediaQuery(theme.breakpoints.down('md'))
-  const { t } = useTranslation('header')
+  const { t } = useTranslation('ui')
```
frzyc, line 151: I think I would prefer to have the `header` be specifically for the header here. I think `ui` might become a shared translation namespace for translations that are done for both SRO and GO.

frzyc (commented):
Minor comment, rest LGTM

frzyc (approved):

PR #2317, opened 2024-07-15
[SRO] Create basic Stat Filter

### LINKED ISSUE #2308: Add min/max stat filter UI + database shape
we need a shape for the min/max stat filter object (we probably dont need to store it in DB yet because none of the loadout handling stuff is added)
We also need a UI for how to specify it. Theres a few avenues we can go down for that
one is like fribbels with a table-esque view. very intuitive
THe other is something similar to GO, where user selects a stat, selects min/max, then inputs the value. With optional checkboxes to toggle the state of the filter
The benefit of that is it takes up less space for unused filters, and allows people to quickly toggle multiple filters

image

tiffanynwyeung:
## Describe your changes
- create basic stat filter UI
- add and modify `StatFilterSetting` to now have `minValue` and `maxValue`
- add `StatFilterCard` to temp optimize page for viewing purposes

## Issue or discord link
* Resolve #2308 

## Testing/validation
Appearance when `targetSelector` on left, with checkbox/delete/input enabled:
image

Alternative layout with `targetSelector` in the middle, with input enabled and disabled respectively:
image
image

### DIFF
--- a/apps/sr-frontend/src/app/Optimize.tsx
+++ b/apps/sr-frontend/src/app/Optimize.tsx
@@ -7,6 +7,7 @@ import { MAX_BUILDS, Solver } from '@genshin-optimizer/sr/solver'
 import {
   BuildDisplay,
   OptimizationTargetSelector,
+  StatFilterCard,
   WorkerSelector,
   useCalcContext,
   useDatabaseContext,
@@ -103,6 +104,7 @@ export default function Optimize() {
     <Container>
       <CardThemed bgt="dark">
         <CardContent>
+          <StatFilterCard />
           <Stack>
             <Typography variant="h5">{t('optimize')}</Typography>
             <Box>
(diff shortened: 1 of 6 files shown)


nguyentvan7 (commented):
Looks good to me, we can refine later as we plan out the opt-tab anyways. I think only comment is maybe we should put the min filter on the left side of the stat select? or is that too clunky/interrupts the left-to-right flow?

tiffanynwyeung:
I've added the alternative layout with `targetSelector` in the center with `min` on the left. I think the flow may be a bit awkward with the checkbox on the left, and there may be potential accessibility issues with needing to select the `targetSelector` first before being able to change input in `min` and `max`? Outside of the checkbox on the left, the middle selector does look better with the current `buttonGroup` design.

I also am not too sure about how to set `maxValue` when there is no user input, from looking at how `StatFilter` in GO gets parsed. It's probably a "putting a cart before the horse" thought exercise, so I decided on setting it to `Infinity` for now, and the logic can be revised when properly implementing opt-tab later. I did come up with the following though: 
- current solution: set `max` to `Infinity`
- making `min` and `max` optional properties that can be removed and added
- possibly accepting `undefined` for `min` and `max` as the default 'no input' value, and later opt-tab code filters out `max/min = undefined`

nguyentvan7 (approved):
Looks good for now thanks!

PR #2343, opened 2024-07-27
Add build/deployment workflows for Somnia

nguyentvan7:
## Describe your changes

* Adjust gi-assets-data and sr-assets-data to export their AssetData without a strict reliance on `dm` modules, which incurs lengthy cold start delays. Now these libraries are similar to gi-stats and sr-stats, where a gen-file executor creates a JSON, which is then imported and exported properly typed. Thus, removing a hard dependency on `dm`.
* Disable databank command for Somnia, as it is not using the i18n library, and thus having a direct reference on the repo, rather than being built into the final package.
* Add build workflow for Somnia that runs on PRs touching certain paths.
* Add deploy workflow similar to our frontend workflows for Somnia. The build workflow will generate an artifact, which will determine if Somnia should even be deployed, as we only have one somnia-test deployment slot, so we don't want to overwrite legitimate testing.

* Deploy workflow has support for 2 different environments: somnia-prod and somnia-test.
* somnia-prod will be deployed to on every push to master
* somnia-test will be deployed to for every PR with an available Somnia build, and also has the label `deploySomnia`
* Both of these bots are already present on the server. We can later limit the permission scope of somnia-test to only be usable in our bot testing channel/thread, so users can use somnia-prod without being confused.

To actually utilize the deployment step, we will need to set up 2 environments on the repo, one for somnia-prod and one for somnia-test.  
Instructions here: https://github.com/frzyc/genshin-optimizer/wiki/Setting-up-repo-environments-for-Somnia-deployments

## Issue or discord link

- 

## Testing/validation

"Successful" deploy - This deployment looks like it failed, but it actually succeeded on Azure. This is a known bug (https://github.com/Azure/functions-action/issues/185). https://github.com/nguyentvan7/genshin-optimizer/actions/runs/10121327262/job/27992122624

Successful build - https://github.com/nguyentvan7/genshin-optimizer/actions/runs/10121389059/job/27992267981

Bot itself works once deployed ~~(tho it does take a few minutes to spin up, and seems prone to flaky startup, also a known issue https://learn.microsoft.com/en-us/answers/questions/168746/container-didnt-respond-to-http-pings-on-port-8080)~~ jk turns out this was caused by bad bot setup, which I fixed.


frzyc (approved):
Looks like we are leveraging every single lesson we learned for CI. Props to connecting all these together.

frzyc (approved):

PR #2371, opened 2024-08-08
Port Translate and Theme to common lib

tiffanynwyeung:
## Describe your changes
- Move `Translate` logic into new generic `TranslateBase` living in `common/ui`, with `Translate` now acting as a wrapper function to call it in both GO and SRO
- Remove `TransWrapper` component (nothing uses it at all across GO/SRO currently)
- Move all shared theme styles to `common/ui`, renaming `brand[GO|SRO]500` to `brand500`
- Generate new SRO theme library (necessary as SRO/`Translate` depends on theme colors)
- Remove extra `Theme` file from `sr-frontend` and call `theme` lib version instead
- Add new yarn cache packages for macOS

Note: current SRO elemental colors are taken from the HSR fandom wiki as placeholders until official colors are established

## Testing/validation
Run `nx serve frontend` and check that translations in different languages are rendering correctly. Some good pages are GO artifact card UI and character talent sheets.


libs/sr/theme/src/Theme.tsx
```diff
@@ -0,0 +1,259 @@
+import { createTheme } from '@mui/material'
+
+import { theme as commonTheme } from '@genshin-optimizer/common/ui'
+
+declare module '@mui/material/styles' {
+  interface Palette {
+    brandSRO500: Palette['primary']
```
frzyc, line 7: brand colors and neutral colors should actually go into ui/common theme

libs/sr/theme/src/Theme.tsx
```diff
@@ -0,0 +1,259 @@
...
+    neutral200: Palette['primary']
+    neutral300: Palette['primary']
+    neutral400: Palette['primary']
+    neutral500: Palette['primary']
+    neutral600: Palette['primary']
+    neutral700: Palette['primary']
+    neutral800: Palette['primary']
+    neutral900: Palette['primary']
+
+    roll1: Palette['primary']
```
frzyc, line 18: roll colors and rarity colors may not make sense in SRO, I would recommend removing them from SRO theme for now, and we can add them later once we need them.

libs/common/ui/src/components/Translate.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { Skeleton, Typography } from '@mui/material'
+import type { TFunction } from 'i18next'
+import type { ReactNode } from 'react'
+import { Suspense } from 'react'
+import { Trans, useTranslation } from 'react-i18next'
+
+/**
+ * Note: Trans.values & Trans.components wont work together...
+ */
+export function TranslateObj({
```
frzyc, line 12: Do you mind renaming this to TranslateBase?

frzyc (dismissed):
Some minor changes

libs/sr/theme/src/Theme.tsx
frzyc: I guess technically roll colours do make sense here, since in SRO there is 1 base roll, and 5 additional rolls, it would make sense to have up to roll6. I would then advocate to have it be separated from GO, and be kept in SRO for now.

libs/sr/theme/src/Theme.tsx
tiffanynwyeung: I noticed that this also exists in GO as a standalone theme library, hence why I generated one - should I also extract these values from that side as well then? I believe almost everything would be inherited from `common/ui` between the two except elemental dmg-related colors, and seemingly roll colors if we choose to make roll colors different for SRO specifically.

frzyc: I think its safe to think that we will share the same theme and colors assumptions in both SRO and GO, so i would agree that both can be moved to common/ui. If we have separate designs for SRO with separate theme, we'd extract it out again.

nguyentvan7 (approved):

PR #2377, opened 2024-08-14
Enforce read-side and write-side query structure

lantua:
## Describe your changes

The current tag db is too lenient on how the buffs get added to a formula. Some formulas are added to `et:self` while others to `et:selfBuff`. This makes the system hard to design as it needs to accommodate both possibilities. This PR enforces more structure on `et:` as follows.

On the read side:
- Final query (value that includes all applicable contributions) reads from `et:self`,
  - `self.final.atk` yields the `final.atk` from character/weapon/artifact/etc,
- Contribution from a particular sheet reads from `et:*Buff` with appropriate `sheet:`,
  - `selfBuff.final.atk.sheet("Nahida")` yields Nahida's contribution to `self.final.atk`.

On the write side:
- Queries with `sheet:undefined` or `sheet:static` write to `et:self`,
- Queries in a sheet write to `et:selfBuff`/`teamBuff`/`notSelfBuff` (i.e., `et:*Buff`),
- Queries outside of a sheet never write to `et:*Buff`,

This structure allows the buff system to be design as
```
`self:` stat <= "determine contributions" <= `et:*Buff` entries
```
It however requires linkage from `self:` <= `et:*Buff` at all time, so we also add `noTeamData` in case there is no team member (and so `src:dst:` are unnecessary).

## Issue or discord link

https://discord.com/channels/785153694478893126/1189329506842984570/1272737073522278570

## Testing/validation

Unit tests are added to enforce write-side entries.

### DIFF
--- a/libs/gi/formula/src/calculator.ts
+++ b/libs/gi/formula/src/calculator.ts
@@ -65,11 +65,11 @@ export class Calculator extends Base<CalcMeta> {
       case 'min':
       case 'max':
       case 'sumfrac': {
-        const empty = arithmetic[op]([], ex)
-        const ops = x.filter((x) => x!.val !== empty) as CalcResult<
-          number,
-          CalcMeta
-        >[]
+        let ops = x as CalcResult<number, CalcMeta>[]
+        if (ops.length > 1) {
+          const empty = arithmetic[op]([], ex)
+          ops = ops.filter((x) => x!.val !== empty)
+        }
         if (ops.length <= 1) {
           let meta = ops[0]?.meta ?? constOverride()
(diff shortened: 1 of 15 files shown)


nguyentvan7 (approved):
Possibly time to rename `selfBuff`? not sure if that name properly describes what it does anymore

Do we need any changes on calc consumption side for this? I'm thinking not atm since we were always reading from `self` for individual char/wep contributions

lantua:
One change to consumption is that individual contribution is read from `*buff`, instead of the previously `self`. If we need contribution from all `selfBuff/teamBuff/notSelfBuff` together, I can add a new read-side `et:` too.

PR #2379, opened 2024-08-18
Add resonance and prepare for buff marking

lantua:
## Describe your changes

This is a hotchpot of several changes related to Pando and gi/sr-formula.

The previous `Calculator` has a problem detecting buff entries. This is because the buff signature (`et:*Buff`) are only available during the gather part of the `read` operation, and are promptly shadowed by the read tag (usually `et:self`). This PR addresses this by adding a callback function `markGathered`, called for each entry found during gathered operation, where `et:*Buff` tags are still intact. (The actual buff marking is for future PRs)

This PR also fixes several bugs in `DebugCalculator`, including
- Preventing crashes when a `unique` read matches zero or multiple entries
- Fixing a bug where entry debug information is missing even when `tag_db` debug flag is set
  - Preventing crashes when such debug information is missing for any other reasons

On the `et:` read/write structure, there are two main changes. Firstly, it connects the linkage between `enemy <- enemyDebuff`, which was previously unconnected, requiring a separate `enemyDebuff` and `enemy` tags. Secondly, `selfBuff` tag is removed. The `selfBuff` tag proves more troubled than its worth as it requires a buff context (`src:dst:`), which is fulfilled by `teamData`. Meaning that `teamData` is required even in a context without team, e.g., weapon page, leading to `noTeamData`. It also adds an extraneous rule surrounding `selfBuff` and `self`, namely, that `selfBuff`/`self` must be used to write inside/outside of a sheet only. With this PR, all entries to `self` is written using `selfBuff`

Currently, reading conditionals assume that both `src:` and `dst:` are set where it is read. This guarantee is important for its operation because a conditional read with tag `src:X dst:Y` may also match a conditional entry `src:X` (without `dst:`). In that case, if a `src:X` conditional is queried (and value supplied) at any point, one must eschew `src:X dst:Y` conditionals as they will use both (the correct) `src:X dst:Y` and (an extraneous) `src:X`. Setting both `src:dst:` avoid this scenario. This PR further enforce this by ensuring that `src:dst:` are set even at a top-level formula. For top-level formulas, `dst:all` is used.

Finally, it adds resonance buffs and rename `Calculator.get` <- `Calculator.gather`.

## Issue or discord link

n/a

## Testing/validation

`gi-formula` and `sr-formula` tests


libs/gi/formula/src/data/common/conds.ts
```diff
@@ -0,0 +1,8 @@
+import { allBoolConditionals } from '../util'
+
+export const { enemyFrozen } = allBoolConditionals('static', 'both')
```
nguyentvan7, line 3: Why can't these be in resonance.ts?

nguyentvan7:
What's the new intention for self vs selfBuff? is self purely read, selfBuff purely write now?

libs/gi/formula/src/data/common/conds.ts
lantua: It could, but these conditionals are available outside of resonance too, e.g., if some char's team/self buff depends on whether an enemy is frozen.

lantua:
> What's the new intention for self vs selfBuff? is self purely read, selfBuff purely write now?

That's the intension, for now.

libs/gi/formula/src/calculator.ts
```diff
@@ -8,19 +8,19 @@ import {
...
-type CondInfo = Partial<
-  Record<
-    Member,
-    Partial<Record<Member, Partial<Record<Sheet, Record<string, number>>>>>
-  >
->
-export type CalcMeta = {
-  tag: Tag | undefined
+type MemRec<V> = Partial<Record<Member, V>>
+type CondInfo = MemRec<MemRec<Partial<Record<Sheet, Record<string, number>>>>>
```
nguyentvan7, line 15: Copy the comment from below here as well?
```suggestion
// conds[dst][src][sheet][condName] = condValue
type CondInfo = MemRec<MemRec<Partial<Record<Sheet, Record<string, number>>>>>
```

libs/gi/formula/src/data/common/conds.ts
```diff
@@ -0,0 +1,8 @@
...
+export const {
+  hasShield,
+  nearbyDendro1, // Burning, Quicken, Bloom
```
nguyentvan7, line 6: nearbyDendro1 and nearbyDendro2 should be moved to resonance. we can leave the other 2

nguyentvan7 (commented):
2 nits, the rest is over my head :NahidaJuice:

libs/gi/formula/src/data/common/conds.ts
lantua: The names still can't collide with other `static` conditionals (the first argument in `allBoolConds`), even if they're declared on different sheets. Wouldn't it be confusing to put them on different files?

nguyentvan7: prepend name with `reso_` or something then. these are still resonance specific so i wouldnt want to export them from common

libs/gi/formula/src/calculator.ts
lantua: Maybe we should *move* it here

libs/gi/formula/src/data/common/conds.ts
lantua: then we can add to `reso` sheet instead

nguyentvan7 (approved):

Issue #2314, opened 2024-07-14
Add UI assets to existing elements


nguyentvan7:
For example, CharacterCard, RelicCard, LightConeCard
some of the existing team UI

can take heavy inspiration from GO

PR #2416, opened 2024-08-28
[SRO] Add basic relic card UI

tiffanynwyeung:
## Describe your changes
- Add basic relic card UI

Notes: we don't have certain dm strings imported at the moment, like sheet set effects, or the individual names/descriptions of set pieces. However, for the latter text specifically, would we want to prefer highlighting the set name + set effects at the top of the card versus the bottom? Or keep it at the bottom and have the piece name/description at the top like how GO does it?

# Issue or discord link
Part of #2314.

## Testing/validation
image


libs/sr/localization/assets/locales/en/relic.json
```diff
@@ -0,0 +1,10 @@
+{
```
frzyc, line 1: Should be able to add this to the datamine extracted hash data, instead of manually translating this. https://github.com/frzyc/genshin-optimizer/blob/master/libs/sr/dm-localization/src/executors/gen-locale/lib/hashData.ts

frzyc (changes requested):
Minor translation note

libs/sr/localization/assets/locales/en/relic.json
nguyentvan7: To do this, open TextMapEN.json and ctrl-f the string you want, such as `"Planar Sphere"`. There will probably be multiple matches, just choose any and copy the key/hash into hashData.ts

frzyc (approved):
LGTM, can hash out the details later.

PR #2424, opened 2024-08-29
Add probabilistic hit for Yanfei A4

nguyentvan7:
## Describe your changes

It's not really possible to optimize with Yanfei A4 in mind. With this change, a user can add this probabilistic hit for each charged attack they do to get an average damage value out of the A4. Not perfect, but better than nothing.
The probability follows the crit hit mode as well.

## Issue or discord link

- 

## Testing/validation

image


frzyc (approved):
Kind of a ugly ducttape, but it is a limitation of WR that we can address this way.

nguyentvan7:
Tbh I don't see a better way to do this in pando either

PR #2440, opened 2024-09-12
Add option to select multiple characters at a time when setting up a team

### LINKED ISSUE #2272: quick character selection in teammate selection UI
Currently need to add each character to the team one by one. Could implement a "quick select" like in game.

Will likely need to add a separate "quick select" button?
need to add a new character select that allows selecting multiple characters, and in order

arquelion:
## Describe your changes
- Added `CharacterMultiSelectionModal.tsx`, which largely maintains the same UI as the current character selection modal but allows selection of multiple characters at once. This UI highlights currently selected characters with a gold outline and number indicating which team slot they are in or will be placed in. Selected characters will always display at the front of the list regardless of active filters/sorts/searches. Team information is updated only when the UI is closed.

`TeamSetting/index.tsx`
- Added `onSingleSelect` and `onMultiSelect`, which both call `onSelect` as needed to modify team data; slightly modified `onSelect` to be usable by both selection functions
- Added a new "Quick Select" button above the individual team slots that brings up the multi selection modal

## Issue or discord link
- Resolves #2272 

## Testing/validation
General multi-selection:
https://github.com/user-attachments/assets/6ee9253b-d0b9-4e1b-9ad0-ea6d858de8fa

Filter/sort/search behavior:

https://github.com/user-attachments/assets/5a9e3739-f4fd-4a24-90ec-1b6e822a4889


libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
```diff
@@ -0,0 +1,500 @@
...
+            sx={{ p: 0.25, position: 'absolute', zIndex: 2, opacity: 0.7 }}
+            onClick={(_) => {
+              onClose()
+              database.charMeta.set(characterKey, { favorite: !favorite })
+            }}
+          >
+            {favorite ? <FavoriteIcon /> : <FavoriteBorderIcon />}
+          </IconButton>
+          {isSelected && (
+            <Typography variant="body2" sx={{ flexGrow: 1 }}>
```
nguyentvan7, line 400: This doesnt seem to update when you 
hover a character to see the tooltip
click
dont move the mouse

it will only start showing/hiding the border once i move my mouse away and it closes the tooltip

nguyentvan7:
this looks super good

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
arquelion: Yeah, I noticed that locally too and wanted to make sure it was also happening in a live environment. Still need to figure out what's causing it, probably something's not updating when I think it is.

frzyc: yeah, its very common to be in a state where the select is not showing 
image

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
```diff
@@ -0,0 +1,500 @@
...
+})
+
+function SelectionCard({
+  characterKey,
+  onClick,
+  selectedCharKeys,
+}: {
+  characterKey: CharacterKey
+  onClick: () => void
+  selectedCharKeys: (CharacterKey | '')[]
```
frzyc, line 347: I would honestly try to simplify this logic by passing in the `selectedIndex: null|number`, instead of the list.

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
```diff
@@ -0,0 +1,500 @@
...
+  const database = useDatabase()
+  const [state, setState] = useState(() => database.displayCharacter.get())
+  useEffect(
+    () => database.displayCharacter.follow((r, s) => setState(s)),
+    [database, setState]
+  )
+
+  const team = database.teams.get(teamId)!
+  const { loadoutData } = team
+  const currentTeamCharKeys : (CharacterKey | '')[] = useMemo(() => {
```
frzyc, line 114: This is the main reason that the child components are not updating:
image
 because this object always has the same reference across renders, which does not trigger rerender in the child components. 
 
 Overall, because people will toggle the characters very fast, if you wait for `loadoutData` to come back, it would requires the whole 4 characters in the team to be recalced, which is very expensive/laggy.
 
 My recommendation would be to store the characters locally in a `const [teamCharKeys, setTeamCharKeys] = useState([])` and only push the change to the selection when you close the modal (I know you are doing half of this already)

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
```diff
@@ -0,0 +1,500 @@
...
+  useEffect(
+    () => database.charMeta.followAny(() => forceUpdate()),
+    [forceUpdate, database]
+  )
+
+  const [searchTerm, setSearchTerm] = useState('')
+  const deferredSearchTerm = useDeferredValue(searchTerm)
+  const deferredState = useDeferredValue(state)
+  const deferredDbDirty = useDeferredValue(dbDirty)
+  const characterKeyList = useMemo(() => {
```
frzyc, line 136: I think you need to keep the selected characters to be displayed at all times, even if there is a filter/search

frzyc (commented):
Some preliminary comments

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
frzyc: This is sort of how I would write this (very roughly)
```Typescript
const [teamCharKeys, setTeamCharKeys] = useState([null,null,null,null])
const { loadoutData } = useTeam(teamId)
// update teamCharKeys when loadoutData changes
useEffect(()=>setTEamCharKeys(loadoutData.map((loadoutDatum) =>
	database.teamChars.get(loadoutDatum?.teamCharId)?.key ?? null
)),[setTeamCharKeys, loadoutData])
```

arquelion:
@frzyc @nguyentvan7 The issue with the selection borders not updating without moving the mouse away should be fixed now, and selected characters now display at the front of the list and should bypass all filters/searches.

The side effect of how I implemented filter bypass is that the character selection cards now automatically reorganize themselves on select/deselect. I played around with it a bit and think I actually prefer it this way, but if you guys think it's a little awkward then let me know.

frzyc:
> @frzyc @nguyentvan7 The issue with the selection borders not updating without moving the mouse away should be fixed now, and selected characters now display at the front of the list when using quick select and should bypass all filters/searches.
> 
> The side effect of how I implemented filter bypass is that the character selection cards now automatically reorganize themselves on select/deselect. I played around with it a bit and think I actually prefer it this way, but if you guys think it's a little awkward then let me know.

I just tried it out. Its a lot more responsive, and all the labels show up properly. 
I do like the fact that selected characters are always shown, regardless of filtering.
I think I would prefer if selection does NOT reorganize the grid. I think the grid should only be reorganized when the filter/sort/search is changed(user is clicking outside of the grid), and not selection(user is clicking inside of the grid). I tend to want to avoid designs where the user selection/view reorganize under themselves on click.(I know we do it for favorites, but I honestly did not like the feel of that either, but that is a separate issue)

arquelion:
> I think I would prefer if selection does NOT reorganize the grid. I think the grid should only be reorganized when the filter/sort/search is changed(user is clicking outside of the grid), and not teammate selection(user is clicking inside of the grid).

Should function like this now, not entirely sure if my solution is the best approach but couldn't think of anything better that worked.

I'd still like to look into merging the single and multi select modals into one file (or at least doing some refactoring on them), though less sure whether it's worth also mimicking in-game behavior in the single selection UI of pinning currently selected teammates to the front but without the slot numbers. I'm also fine with getting this PR cleaned up and ready to be merged and opening a new PR for refactoring and any related changes to single selection.

frzyc:
Interactions in the quick select modal is working properly as expected, and feels nice.
Closing the modal via clicking outside the modal transfer the list to the team
However, closing the modal via clicking the "X" on the top right does not transfer the team.

In terms of refactoring, I'd recommend starting a new PR, so we can roll out this feature for people to use first. 

I like your idea of "outline the selected character when you reopen the modal", definitely worth delving into in the future.

libs/common/ui/src/theme/index.ts
```diff
@@ -41,6 +41,8 @@ declare module '@mui/material/styles' {
     rarity3: Palette['primary']
     rarity4: Palette['primary']
     rarity5: Palette['primary']
+
+    charSelected: Palette['primary']
```
frzyc, line 45: I am not sure if we need to have a separate color for this. I'd say we should use the default action color palette (success/warning/failure/info). 
In terms of color, I would actually recommend you to use "success"(greenish) instead of this yellow color, see how that looks.

arquelion: image
I prefer the gold outlines, personally. The green isn't as visually distinct against the modal's background color and blends in with some of the namecards as well.

frzyc: I see. I am fine keeping it yellow, or warning for now

arquelion:
Actually, should multi-select have a confirm button instead of always updating as soon as it's closed? I worry it'll be very annoying if someone makes some changes to their team selection in the UI but then decides to not keep those changes; those changes will end up being saved as soon as the UI closes and they'll have to manually undo them. With the button, only clicking on confirm will save the new team selections while all other methods of closing the UI will discard them.

Just a passing thought I had, dunno how often this particular use case will come up but maybe a change worth making for convenience reasons.

frzyc:
> Actually, should multi-select have a confirm button instead of always updating as soon as it's closed? I worry it'll be very annoying if someone makes some changes to their team selection in the UI but then decides to not keep those changes; those changes will end up being saved as soon as the UI closes and they'll have to manually undo them. With the button, only clicking on confirm will save the new team selections while all other methods of closing the UI will discard them.
> 
> Just a passing thought I had, dunno how often this particular use case will come up but maybe a change worth making for convenience reasons.

I guess i prefer to cater and minimize the main usecase, which is a user selecting the teammates as they intended, without having to click a "confirm"(having to confirm team changes frustrate me in genshin as well)

arquelion:
small update: I've made the last couple changes (heading + smaller quick select button, fix for team data not updating when using the X button to close the modal) locally but ran into probably an edge case while grabbing validation videos. Should be good for a final review as soon as I can figure out what's causing it.

arquelion:
@frzyc Did not find any more bugs while grabbing validation videos, ready for final review

libs/gi/page-team/src/TeamSetting/index.tsx
```diff
@@ -191,6 +249,14 @@ function TeamEditor({
           character" buffs, and cannot be empty.
         </Trans>
       </Alert>
+      <Grid container sx={{ justifyContent: 'space-between' }} spacing={2}>
+        <Grid item>
+          <Typography variant="h5">Selected Team Members</Typography>
+        </Grid>
+        <Grid item>
+          <Button onClick={() => setShowMultiSelect(true)}>Quick Select</Button>
```
frzyc, line 257: Maybe add a icon to the button, I would recommend https://mui.com/material-ui/material-icons/?query=multiple&selected=Filter4

frzyc (changes requested):
Mostly LGTM. Very minor changes requested.
I understand the ask of "toggling teammates do not change the list" might be a complicated one to implement, your solution is nicely done.

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
```diff
@@ -0,0 +1,523 @@
...
+}: CharacterMultiSelectionModalProps) {
+  const { t } = useTranslation([
+    'page_character',
+    // Always load these 2 so character names are loaded for searching/sorting
+    'sillyWisher_charNames',
+    'charNames_gen',
+  ])
+  const { silly } = useContext(SillyContext)
+  const database = useDatabase()
+  const [state, setState] = useState(() => database.displayCharacter.get())
```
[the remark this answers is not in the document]
arquelion: also changed the same line in `CharacterSelectionModal` since this was copied from there

libs/gi/ui/src/components/character/CharacterMultiSelectionModal.tsx
```diff
@@ -0,0 +1,523 @@
...
+  const deferredState = useDeferredValue(state)
+  const deferredDbDirty = useDeferredValue(dbDirty)
+  const characterKeyList = useMemo(() => {
+    const { element, weaponType, sortType, ascending } = deferredState
+    const sortByKeys = [
+      ...(newFirst ? ['new'] : []),
+      ...(characterSortMap[sortType] ?? []),
+    ] as CharacterSortKey[]
+    if (deferredDbDirty) {
+      const filteredKeys = allCharacterKeys
```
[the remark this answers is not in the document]
arquelion: done

libs/gi/page-team/src/TeamSetting/index.tsx
arquelion: done

frzyc (approved):
LGTM. I can see this this was a sizable PR, I appreciate you for being resilient and sticking with it.

libs/gi/page-team/src/TeamSetting/index.tsx
```diff
@@ -191,6 +250,19 @@ function TeamEditor({
           character" buffs, and cannot be empty.
         </Trans>
       </Alert>
+      <Grid container sx={{ justifyContent: 'space-between' }} spacing={2}>
+        <Grid item>
+          <Typography variant="h5">Selected Team Members</Typography>
```
frzyc, line 255: Opps, forgot to remind you to i10n these newly added text. I think In this line, "Team Members" would be more appropriate, and for the button, "Quick setup" can be used.
image

frzyc: perhaps we can ~~steal~~ borrow the i10n from datamine for "Quick Setup"
image

PR #2442, opened 2024-09-15
Fix couple stat issues

nguyentvan7:
## Describe your changes

* Remove -1 from things that have no levels
* Adjust char dm util to reduce nesting needed to access values
* Adjust heal util for future heals that don't have a flat value
* Rename brEff_ -> brEffect_ to avoid collision with brEfficiency_ later

## Issue or discord link

- 


frzyc (approved):
I am okay with the stat renaming, just need to make sure scanners continue to work. (do we need to raise an issue with scanner devs?)

nguyentvan7:
Scanners been broke plus we don't even have support from reliquary scanner which is the primo scanner imo. Might be better for us to support their format instead of other way around maybe?

frzyc:
> Scanners been broke plus we don't even have support from reliquary scanner which is the primo scanner imo. Might be better for us to support their format instead of other way around maybe?

sure, we'd just add a conversion layer between.

PR #2460, opened 2024-09-25
Standardize scalesWith across character UI

frzyc:
## Describe your changes

Hoist up a "scales with" list into a context to be used across Character UI (TC opt, opt, overview etc)
Add a "Scales with" alert to the optimization page
Remove the build constraint in TC builds, opting to use the build constraints from optimization page.

## Issue or discord link

- 

## Testing/validation
Screenshot 2024-09-25 164149


nguyentvan7:
If I toggle spread on and off, it changes the order of the stats. Should we deterministically sort this to avoid the implication that one stat is weighted more? I might be overthinking it

image
image

nguyentvan7:
Technically ER should be here, as I only enabled EOSF, and in general EOSF does just make burst scale off ER, but i guess not really known unless the character is currently equipping EOSF

image

e.g. when char has EOSF equipped, ER shows up
image

libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/ScalesWith.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Alert } from '@mui/material'
+import { useContext } from 'react'
+import { Trans, useTranslation } from 'react-i18next'
+
+export default function ScalesWith() {
+  const { t } = useTranslation('page_character_optimize')
+  const { scalesWith } = useContext(OptTargetContext)
+  return (
+    !!scalesWith.size && (
+      <Alert severity="info" variant="filled">
```
nguyentvan7, line 16: Can we use the new standard for alert?
image

frzyc: you meant the `variant="standard"`? I'll give it a try.

frzyc:
> Technically ER should be here, as I only enabled EOSF, and in general EOSF does just make burst scale off ER, but i guess not really known unless the character is currently equipping EOSF
> 
> image
> 
> e.g. when char has EOSF equipped, ER shows up image

Yeah that is an issue that would be pretty difficult to address, Another cornercase would be that if you currently have ESOF equipped, and it generates a build without ESOF, it would stills show it scaling with ER. I am honestly not sure if I want to commit the time to solve this since any solution I can think of involves a lot of spaghetti.

frzyc:
> If I toggle spread on and off, it changes the order of the stats. Should we deterministically sort this to avoid the implication that one stat is weighted more? I might be overthinking it
> 
> image image

Random order would actually avoid the implication that there is a priority, no? Also yeah its not ranked (because how can we determine that even) so its not really an issue worth addressing.

nguyentvan7 (approved):

PR #2471, opened 2024-09-30
[SRO] Add LC translation strings & LC card UI

tiffanynwyeung:
## Describe your changes
- create proper `LightConeCard` UI
- create string translations for light cones and add them to card and editor
- fix `LightConeEditor` to be able to edit/update existing light cones properly
- fix Erudition path svgicon to no longer have invalid path code embedded inside
- fix relic localization strings to reference `Relic` instead of GO `Art`
- extreme style/nit: change from mixing translated string calls as `` t`string` `` to strictly using `t('string')` in LC/relic

Notes:
- Superimposition TL string does not use `<1>{value}</1>` to add `<strong>1/2/3/4/5</strong>` on the cards anymore: the original is not reusable with the `SuperimposeDropdown` component, and the visual clarity from bolding the number seems minimal from inspection that we could simply make the TL string more reusable
- "Lv." TL string from DM has `{SPACE}` keyword in certain languages, and currently no processing exists to parse it out at the moment. Needs to be fixed in future PR
- `LightConeEditor` is still following `RelicEditor` closely in terms of UI: currently, there's no way to load light cone sheets as UI data yet. Not sure if adding this ability to edit existing light cones in the manner it's written right now will be worth since it's handled differently in GO, and if we want exact parity with GO on this case, it'll probably end up being removed in a later PR
- Still missing ways to save DB UI settings in SRO so we can add sort/filter options to relics/LCs

## Issue or discord link
Part of resolving #2314. 

## Testing/validation
New light cone card UI
image

Editing existing light cone
image

Light cone inventory after editing Bronya LC from S3 -> S1 and Lv. 60/70 -> Lv. 80/80
image


frzyc:
> extreme style/nit: change from mixing translated string calls as `` t`string` `` to strictly using `t('string')` in LC/relic

@tiffanynwyeung Can you elaborate why you prefer the latter style? I would honestly prefer the template literal notation for translations, its more compact.

libs/sr/localization/assets/locales/en/lightCone.json
```diff
@@ -0,0 +1,41 @@
...
+    "btnAdd": "Add Light Cone",
+    "btnClear": "Clear",
+    "btnRandom": "Randomize",
+    "btnUpdate": "Update Light Cone",
+    "clearPrompt": "There is a light cone being edited. Are you sure you want to clear the editor?",
+    "delete": "Delete",
+    "confirmDelete": "Are you sure you want to delete this light cone?"
+  },
+  "level": "Lv.",
+  "superimpose": "Superimposition {{rank}}",
```
frzyc, line 19: Can use dm for l10n
"-2008197403": "Superimposition",

libs/sr/localization/assets/locales/en/lightCone.json
```diff
@@ -0,0 +1,41 @@
...
+    "title": "Light Cone Editor",
+    "unknownLightCone": "Unknown Light Cone",
+    "lightConeName": "Light Cone Name",
+    "levelSelect": "Select Level",
+    "preview": "Light Cone Preview",
+    "beforeEdit": "Before Edit",
+    "btnSave": "Save Light Cone",
+    "btnAdd": "Add Light Cone",
+    "btnClear": "Clear",
+    "btnRandom": "Randomize",
```
frzyc, line 12: Clear + Random should go into the `common` namespace https://github.com/frzyc/genshin-optimizer/blob/master/libs/common/localization/assets/locales/en/common.json

nguyentvan7:
> > extreme style/nit: change from mixing translated string calls as `` t`string` `` to strictly using `t('string')` in LC/relic
> 
> @tiffanynwyeung Can you elaborate why you prefer the latter style? I would honestly prefer the template literal notation for translations, its more compact.

I've never seen us use the tag template notation for anything the function notation doesn't cover. And function notation is more well known imo. Just my 2c

tiffanynwyeung:
@frzyc 
> There are existing `Lv.` translations in DM as well, I am not sure if we have handling in our pipeline to handle the `{SPACE}` syntax

Good catch on that + the other DM string suggestions, I keep forgetting that we can use DM translations out of there and I should make sure to remember to use that option more often. As far as I can see, we do have `Lv` already extracted in `sheetHashData` but not used anywhere at the moment, but I also can't find any processing to extract any of the `{SPACE}` syntax in certain langs in that file or anywhere else. I could try and look into how to do this in another PR after switching to those DM strings in this one?

> Can you elaborate why you prefer the latter style? I would honestly prefer the template literal notation for translations, its more compact.

To piggyback on what @nguyentvan7 added, outside of function notation being more well-known, we already fall back to using it when needing to load certain values into a translation, e.g. in `SubstatInput` component we have this line: 
https://github.com/frzyc/genshin-optimizer/blob/3a570923699d19499a0f27b1558edc1a0dab4ea9/libs/gi/ui/src/components/artifact/editor/SubstatInput.tsx#L78
which seems necessary as the TL string requires a custom number value passed in. (And I assume we'll have other similar cases as we port more features.)

My preference is if we could make notation usage more consistent overall. Since we use function notation by default in those cases, I thought it would be better to use that as the standard even with static strings. I'm also open to reverting this style commit and preemptively grouped most of those changes as such if we want to move this to a separate PR/overlook this style refactor, but my initial assumption was that this would seem somewhat excessive as its own PR when we could introduce it in small, iterative commits. Open to feedback about any of this, of course!

frzyc:
You are right, they are already translated in sheet_gen.json, you can use them directly:
image

I guess I don't have as much reservations towards the notation direction, as long as we are consistent. 
Currently, in the repo, there are 426 instance of `` t` ``, that you will have to eventually address if you prefer the functional style (i think we can do it via a simple regex search and replace)

tiffanynwyeung:
Sounds good, I'll do the notation regex-replace in a future PR, but leave the current notation changes in for now

frzyc (approved):
LGTM, thanks for the discussions.

PR #2473, opened 2024-10-01
CharacterSelectionModal refactoring and followup changes for single selection

arquelion:
## Describe your changes
`CharacterSelectionModal` and `CharacterMultiSelectionModal` are now in the same file and refactored as follows:
- `CharacterSingleSelectionModal` and `CharacterMultiSelectionModal` are the exported components utilizing the rest of the components in `CharacterSelectionModal.tsx` to form the modal's structure and interactivity. These modals no longer fetch `loadoutData` themselves and instead have them passed in as props when necessary.
- `CharacterSelectionModalBase` contains most of the structural and filter/search/sort code. `CharacterSingleSelectionModal` and `CharacterMultiSelectionModal` are responsible for passing in the vast majority of functions, character key lists, etc. that this component needs (listed in `CharacterSelectionModalBaseProps` and `FilterSearchSortProps`).
- `SelectionCard` has had `CustomTooltip`, `CardThemed`, and the `IconButton` for favorites moved into new components `SingleSelectCardWrapper` and `MultiSelectCardWrapper`. `SingleSelectCardWrapper` also contains the new functionality for animating the outline on the slot currently being selected, and `MultiSelectCardWrapper` contains the functionality added in PR 2440 for outlining currently selected teammates and indicating their team slots for multi selection.
- `CharacterMultiSelectionModal.tsx` deleted as a result of these changes and all current call sites changed to use single or multi select as necessary

Single character selection in `CharacterSelectionModal.tsx`:
- Currently selected teammates (when relevant) are now pinned to the front of the selection list and bypass filter/sort/search
- Instead of a chip indicating the character's current team slot, single selection simply outlines currently selected characters but flashes the outline for the slot being changed. (Outlines transition between gold and the background color of the modal as this is significantly more visible than flashing white.)

## Issue or discord link
- Follow-up work for #2440 

## Testing/validation
Team single select outlines:

https://github.com/user-attachments/assets/b5e523ce-f352-48c4-a9b2-a49ee15acf8c

I've manually checked all interactions I can think to confirm the refactor has not broken anything.


frzyc (commented):
Before you go too far with this, I do not think it is a good idea to merge the logic of 2 elements with two different usecases, since that makes the core code path more complicated and harder to maintain. If you want to reduce duplication of code, hoist out the common elements into their separate function/element.

arquelion:
Honestly I was kind of wondering whether to split this back up into a regular `CharacterSelectionModal` and a `TeamCharacterSelectionModal`. With the changes to single select to display all currently selected team members etc., single and multi select where teams are concerned actually share almost everything, just a couple small differences in whether to show the team slot number and whether to flash the outline for the currently selected slot. It does feel like there's just enough different in the code between the Teams use case and the Characters page use case though.

frzyc:
Yeah I understand your concern, In this case, I think you should still have some division of components:

## SingleSelect(4 selected, currentlySelected, onSelect)
- CardWrapper(flashing/selected +index)
  - characterCard
  
  
## MultiSelect(4 selected, onSelectTeam) 
store 4 selected state locally(for quick selection + save)
- CardWrapper(show border+index)
  - characterCard

This should make sure that each component only does one thing with one unique state each, instead of a super component that does 2 things in different states depending on usecase

arquelion:
Trying one more approach based on your suggestion before I throw in the towel and deal with all the duplicated code: I split out some of the shared modal code out into a `CharacterSelectionModalBase` that both the single and multi select versions use and also split out some of the `SelectionCard` code into the two wrapper components `SingleSelectCardWrapper` and `MultiSelectCardWrapper` which also contain their respective outlining/animation/slot indicators (so `SelectionCard` really is just the card now).

arquelion:
@frzyc I'll go through it again to see what could be cleaned up, but otherwise good for a first round of review

frzyc:
@arquelion This looks good to me. I like the organization and the separation of concerns here.

arquelion:
Okay, I went through and added some usage comments but otherwise didn't find anything else that I felt needed comments that didn't already have them. Also added a quick recording of the new additions to single select and manually checked as many interactions as I could think of. (Second pair of eyes double-checking all the possible use cases would be nice though.)

frzyc (approved):
LGTM, Thanks for iterating on this.

PR #2485, opened 2024-10-07
pipe stat filter to solver

frzyc:
## Issue or discord link

- 


libs/sr/solver/src/childWorker.ts
```diff
@@ -120,17 +124,23 @@ async function start() {
...
-                ids: {
-                  head: head.id,
-                  hands: hands.id,
-                  feet: feet.id,
-                  body: body.id,
-                  sphere: sphere.id,
-                  rope: rope.id,
-                },
-              })
+              if (
```
nguyentvan7, line 127: This messes up the calculated build count, since that is handled in sliceSortSendBuilds(), which occurs after this

libs/sr/solver/src/childWorker.ts
```diff
@@ -123,7 +123,7 @@ async function start() {
                 sphere.stats,
                 rope.stats,
               ])
-
+              let skipped = 0
```
nguyentvan7, line 126: Isn't this just getting reset every loop iteration?

frzyc: You are right, my bad.

nguyentvan7 (approved):

PR #2491, opened 2024-10-10
[SRO] Organize DM-l10n strings

### LINKED ISSUE #2475: [SRO] DM string organization for gen-files
We want to shift to using DM translated strings as much as possible, though we don't have any current conventions on how those strings could be organized. It would be nice to create some sort of guidelines for that and rename/shuffle current strings while it's still early.

tiffanynwyeung:
## Describe your changes
__**SRO**__
- Reorganize `hashData` namespaces and related files, converting references as necessary
- Remove `{SPACE}` artifact during DM string preprocessing for certain languages
- Experimental: extract non-game-related, generalized UI strings for project use
- Generate new hashdata namespace files

__**GO**__
- Add dm-l10n strings for Team Settings page as requested
- Create new `teams_gen` dm-l10n file

Notes:
- A lot of strings don't have a counterpart, or a perfect singular/plural pair. All strings with both plural and single forms are converted to use i18next format for plurals, but strings missing a plural/single form are referenced normally. 
- UI strings is **experimental** as we need some discussion on if we want to go this route and how to make the pipeline work, as mentioned previously. The file is mostly there to provide a concrete example of what the raw data COULD look like, and/or if we wanted to use it somehow?

## Issue or discord link
Resolves #2475.


libs/sr/dm-localization/src/executors/gen-locale/lib/hashData.ts
```diff
@@ -36,40 +37,6 @@ export const HashData = {
   lightCone: lightConeHashData,
   lightConeNames,
 
-  sheet: sheetHashData,
```
frzyc, line 39: One of the reason why we have separated namespace -> separated .JSON is that translations are usually lazy loaded. when the artifact page is loaded, it doesn't need any of the character/team UI translations, and vice versa. This can impact load time of pages and elements, since most of our elements are behind loading barriers and suspenses. we can designate some of the namespace to be permanently loaded(i think "ui" is).

frzyc:
Also, can you bundling in l10n for these strings https://github.com/frzyc/genshin-optimizer/pull/2440#discussion_r1775930079, I believe they can be partially retrieved from dm.

frzyc:
looks mostly organized for me. for the singular/plural, I think i prefer in the way indicated by the [i18next format](https://www.i18next.com/translation-function/plurals):
```
"character_one": "character"
"character_other": "characters"
```
rather than the `label.one`/`label.many`

frzyc (approved):
LGTM

PR #2507, opened 2024-10-17
Remove submodule checkout for release build

nguyentvan7:
## Describe your changes

It wasn't used since we changed to checking in all generated files
I am not sure why CI gen-file check doesn't fail when doing the sparse checkout of StarRailData. I assume it might be because of the cache being used? But it is possible that might break at some point

## Issue or discord link

- 


frzyc (approved):
lmao 2 line change coulda saved you from manually deploying half a dozen times

nguyentvan7:
It's cause I thought it was a hard requirement for release, but then I remembered we checked in all generated files now, even strings

PR #2511, opened 2024-10-19
Bugfixes for comparisons when editing an inactive build and team multi-select

### LINKED ISSUE #2466: Equip confirmation displays the wrong comparison when modifying an inactive build
The comparisons displayed in the equipment swap confirmation are always for the currently active build, which results in the wrong comparisons being displayed when modifying an inactive build. Only the comparison is incorrect; equipment changes are made to the correct build.

https://github.com/user-attachments/assets/122efac2-abeb-4814-be8a-04c39452c525

arquelion:
## Describe your changes
**Fix for incorrect comparisons being displayed when trying to edit a build other than the current active build**
New file `BuildEditContext.tsx`:
- In order to display the correct comparisons for inactive builds, information regarding which build is currently being edited needs to be piped through to `CompareBuildWrapper` instead of always fetching `LoadoutDatum.buildId`, which always corresponds to the active build if the active build isn't the equipped build. `BuildEditContext` contains a single string set to either `"equipped"` (if the equipped build is being edited) or `LoadoutDatum.buildId` (if a non-equipped build is being edited). `CompareBuildWrapper` will then use this to fetch the weapon/artifact ids accordingly. This avoids needing to manually pipe this value through all the components on the build edit path.
- Export added to `gi/ui/src/context/index.ts`

`TeamDataManager.ts`:
- Added new methods `getEditWeaponId`/`getEditArtifactIds` - These determine the type of build being edited then pass that along with `teamCharId` and the `buildId` to the new private methods below to actually fetch the ids needed for comparison. Build types are deduced to always be either `'equipped'` or `'real'` as it makes no sense to edit a TC build in this context.
- Added private methods `getWeaponId` and `getArtifactIds` - These contain the code previously in `getLoadoutWeaponId` and `getLoadoutArtifactIds` to reduce code duplication.
- Modified `getLoadoutWeaponId`/`getLoadoutArtifactIds` to use the private methods above (functionality unchanged)

`BuildReal.tsx`, `EquipmentSection.tsx`, `character/editor/Content.tsx`:
All three of these locations now wrap their `EquippedGrid` component with `BuildEditContext.Provider` and provide either `buildId` or `"equipped"` as the value.

**Bonus bugfix: Teammate multi-select drops a character if that character was moved to a different slot from the first slot and the team is not full**
`TeamSetting/index.tsx`:
Changed the last conditional in `onSelect` to use `database.teams.get(teamId)` instead of `team` from earlier in the function - This ensures that changes applied so far to the team are accounted for when checking whether a character has been displaced from their old slot regardless of which slot it was, preventing conditional from being false in cases where a character has been moved but `loadoutData[selectedIndex]` is undefined.

## Issue or discord link
- Fixes #2466 
- Fixes this bonus bug discovered while testing the comparison fix

https://github.com/user-attachments/assets/9012aeb2-a5c8-44da-ba56-40d97da91dc1

## Testing/validation
Build edit comparison fix:

https://github.com/user-attachments/assets/7a226a74-74db-4922-a0cd-726e190da4b9

https://github.com/user-attachments/assets/8bb382c1-d58a-4636-8930-0f5e5535f7fa

https://github.com/user-attachments/assets/3f69ba11-673a-454f-b976-16609cf6d64a

Team multi-select fix:

https://github.com/user-attachments/assets/19016ac0-9be2-4645-842d-c830e973b095

### DIFF
--- a/libs/gi/page-team/src/CharacterDisplay/Build/BuildReal.tsx
+++ b/libs/gi/page-team/src/CharacterDisplay/Build/BuildReal.tsx
@@ -8,6 +8,7 @@ import {
   useDatabase,
   useEquippedInTeam,
 } from '@genshin-optimizer/gi/db-ui'
+import { BuildEditContext, type BuildEditContextObj } from '@genshin-optimizer/gi/ui'
 import { getCharStat } from '@genshin-optimizer/gi/stats'
 import {
   ArtifactCardNano,
@@ -27,7 +28,7 @@ import {
   IconButton,
   TextField,
 } from '@mui/material'
-import { useContext, useDeferredValue, useEffect, useState } from 'react'
+import { useContext, useDeferredValue, useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 export default function BuildReal({
@@ -217,10 +218,6 @@ function BuildEditor({
     const { name, description } = newBuild
     setName(name)
     setDesc(description)
-    database.teams.setLoadoutDatum(teamId, teamCharId, {
-      compareType: 'real',
-      compareBuildId: buildId,
-    })
   }, [database, buildId, teamId, teamCharId])
 
   useEffect(() => {
@@ -238,6 +235,11 @@ function BuildEditor({
     // Don't need to trigger when buildId is changed, only when the name is changed.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [database, descDeferred])
+
+  const buildEditContextObj = useMemo(() =>
+    ({buildToEdit: buildId} as BuildEditContextObj)
+  , [buildId])
+
   return (
     <CardThemed>
       <CardHeader
@@ -266,19 +268,21 @@ function BuildEditor({
           minRows={2}
         />
         <Box>
-          <EquippedGrid
-            weaponTypeKey={weaponTypeKey}
-            weaponId={build.weaponId}
-            artifactIds={build.artifactIds}
-            setWeapon={(id: string) =>
-              database.builds.set(buildId, { weaponId: id })
-            }
-            setArtifact={(slotKey, id) =>
-              database.builds.set(buildId, (build) => {
-                build.artifactIds[slotKey] = id ? id : undefined
-              })
-            }
-          />
+          <BuildEditContext.Provider value={buildEditContextObj}>
+            <EquippedGrid
+              weaponTypeKey={weaponTypeKey}
+              weaponId={build.weaponId}
+              artifactIds={build.artifactIds}
+              setWeapon={(id: string) =>
+                database.builds.set(buildId, { weaponId: id })
+              }
+              setArtifact={(slotKey, id) =>
+                database.builds.set(buildId, (build) => {
+                  build.artifactIds[slotKey] = id ? id : undefined
+                })
+              }
+            />
+          </BuildEditContext.Provider>
         </Box>
       </CardContent>
     </CardThemed>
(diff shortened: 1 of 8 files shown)


libs/gi/ui/src/context/BuildEditContext.tsx
```diff
@@ -0,0 +1,12 @@
+'use client'
+import { createContext } from 'react'
+
+export type BuildEditContextObj = {
+  buildToEdit: string
```
frzyc, line 5:
```suggestion
  buildToEdit: string | "equipped"
```

libs/gi/ui/src/context/BuildEditContext.tsx
```diff
@@ -0,0 +1,12 @@
...
+  //setBuildIdToEdit: (buildId: string) => void
+}
+
+export const BuildEditContext = createContext({
```
frzyc, line 9: could simplify this from an obj to an string, and maybe make "equipped" as default?
```suggestion
export const BuildEditContext = createContext<string | "equipped">("equipped")
```

libs/gi/ui/src/components/build/CompareBuildWrapper.tsx
```diff
@@ -25,24 +26,30 @@ export function CompareBuildWrapper(props: WrapperProps) {
...
+    buildToEdit,
+    teamCharId
+  )
   const newWeaponId = weaponId ?? currentWeaponId
   return (
     <EquipBuildModal
       currentName={
-        (loadoutDatum.buildType === 'real' &&
-          database.builds.get(loadoutDatum.buildId)?.name) ||
+        (buildToEdit !== '' && database.builds.get(buildToEdit)?.name) ||
```
frzyc, line 52:
```suggestion
        database.builds.get(buildToEdit)?.name ?? "Equipped"
```

libs/gi/page-team/src/CharacterDisplay/Build/BuildReal.tsx
```diff
@@ -212,7 +230,7 @@ function BuildEditor({
     const { name, description } = newBuild
     setName(name)
     setDesc(description)
-  }, [database, buildId])
+  }, [database, buildId, teamId, teamCharId])
```
frzyc, line 233: I am not sure if adding these to the dependency is necessary, can you explain?

frzyc (changes requested):
Looks mostly good to me. thanks for resolving this difficult issue. I have some minor comments that warrants addressing.

libs/gi/page-team/src/CharacterDisplay/Build/BuildReal.tsx
arquelion: Oh woops, those aren't supposed to be there anymore. Must've snuck back while I was solving the merge conflicts during rebase

libs/gi/ui/src/context/BuildEditContext.tsx
arquelion: done

libs/gi/ui/src/context/BuildEditContext.tsx
arquelion: done

libs/gi/ui/src/components/build/CompareBuildWrapper.tsx
arquelion: done

frzyc (approved):
LGTM, thanks for your work!

PR #2538, opened 2024-11-10
Improve optimization progress accuracy

TornOne:
## Describe your changes
Updated the progress bar when optimizing artifacts to better reflect the actual percentage of progress done.  
The new formula is based on the assumption that the amount of skipped builds remains proportional to the amount of tested builds, but also improves accuracy in cases where more builds are skipped closer to the end than the beginning, and cases where not too many more builds are skipped closer to the beginning than the end.  
In contrast, the previous formula essentially assumed all builds that were going to be skipped were skipped as soon as the generation started.

## Issue or discord link
https://discord.com/channels/785153694478893126/1296108919608971334

## Summary by CodeRabbit

- **New Features**
	- Updated progress calculation in the Build Alert to include skipped builds, providing a more accurate representation of build status.
- **Bug Fixes**
	- Improved display of progress percentage in the Build Alert based on the new calculation method.

### DIFF
--- a/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/Components/BuildAlert.tsx
+++ b/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/Components/BuildAlert.tsx
@@ -55,7 +55,7 @@ export default function BuildAlert({
   let progress = undefined as undefined | number
 
   if (generatingBuilds) {
-    progress = (tested * 100) / unskipped
+    progress = ((tested + skipped) / total) * 100
     title = (
       <Typography>
         Generating and testing {testedString}
--- a/libs/gi/ui/src/components/character/BuildAlert.tsx
+++ b/libs/gi/ui/src/components/character/BuildAlert.tsx
@@ -69,7 +69,7 @@ export function BuildAlert({
   let progress = undefined as undefined | number
 
   if (generatingBuilds) {
-    progress = (tested * 100) / unskipped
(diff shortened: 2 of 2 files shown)


lantua:
What's the behavior for this progress bar if `skipped >> unskipped` since the beginning? We use the current formula because the previous one got "stuck" at ~99% when `skipped` is larger than `unskipped` by several order of magnitudes, which was common at the time (and hopefully is still common nowadays).

TornOne:
Indeed, if the amount of skipped builds makes up for a large majority of the total builds at the beginning, then the bar will spend more time seemingly near completion, as you said.  
I have never seen this happen since some update to the the way the artifacts were processed some years back. I have also tested this is with a "normal" batch of artifacts, and the behavior seems to remain exactly the same, so my unique common usage of GO should not be a factor.

I am curious if you could provide examples of the behavior you describe happening in real-world use, because in my experience, the opposite is somewhat common currently - the bar gets "stuck" at 0%, spending 10-100 times (this depends on how many more builds are skipped than unskipped) more time going from 0% -> 1% than from 99% to 100%. I would hope the change fixes this dramatic "speedup" I regularly see.

lantua:
I couldn't find the old benchmark data, and we don't have a perf test either :D
When I said there would be many prunings at the beginning, I may have been thinking about the "first prune" (one before Solver instantiation), but that shouldn't affect this PR.

A more recent test has around 3x pruning instead, and this PR should work fine enough.

lantua (approved):

Issue #2549, opened 2024-11-20
Locked state detection


frzyc:
Some scanners do not scan in locked state. Can add logic similar to location handling: if there are no location data in the import, do not overwrite location.

location handling: https://github.com/frzyc/genshin-optimizer/blob/master/libs/gi/db/src/Database/DataManagers/ArtifactDataManager.ts#L173

AiZm8:
I would like to work on this issue

PR #2575, opened 2024-12-13
Preserve Locked State for Artifacts

AiZm8:
## Describe your changes

When existing locked artifacts get updated by one that is unlocked, the locked state is preserved

## Issue or discord link

- https://github.com/frzyc/genshin-optimizer/issues/2549

## Testing/validation

I tested it by uploading a database with unlocked artifacts, manually locked them in the site, and then re-uploaded the same JSON batch to see if the artifacts I locked became unlocked or not
Additionally tested for if the lock property is missing from the new JSON upload ("lock": false/true) by removing it, and the artifact remains in its former state (locked/unlocked)

## Summary by CodeRabbit

- **New Features**
	- Enhanced artifact import functionality to preserve the lock state during operations.

- **Bug Fixes**
	- Improved handling of artifact lock states to ensure accuracy during imports.


libs/gi/db/src/Database/DataManagers/ArtifactDataManager.ts
```diff
@@ -215,6 +215,8 @@ export class ArtifactDataManager extends DataManager<
           importArt = {
             ...art,
             location: hasEquipment ? art.location : match.location,
+            // Preserve existing lock state for artifacts reported as unlocked or when the scanner doesn't detect locks ("lock": false JSON)
+            lock: art.lock === false ? match.lock : art.lock,
```
frzyc, line 219: I would preferred if we do 
```
const hasLock = artifacts.some((a)=>a.lock)
...
lock: hasLock? art.lock : match.lock
```
similiar to `hasEquipment`

since this logic would not allow unlocking of artifacts from the scanned data (if user unlocked in game, and scanner scanned it, but it's locked in GO)

frzyc (changes requested):

libs/gi/db/src/Database/DataManagers/ArtifactDataManager.ts
AiZm8: I made the requested change and confirmed with testing where having all the lock properties as false keeps the original lock statuses (locked/unlocked), and if there are locks present, artifacts can become be locked or unlocked

frzyc (approved):

PR #2597, opened 2025-01-16
Fixes: Clear TC autocomplete on selection, Allow commas in number inp…

failchon:
…uts, Allow ctrl+v on artifact editor

## Describe your changes

- Artifact set Autocomplete on TC page will now clear after selecting an artifact set
- CTRL + V should work on the Artifact Editor modal regardless of where you click
- Allow commas in the number inputs.

## Issue or discord link

[- CTRL+V not working](https://discord.com/channels/785153694478893126/1327301891188850779)

## Testing/validation

- None

## Summary by CodeRabbit

- **New Features**
	- Enhanced numeric input handling in the NumberInputLazy component to support more flexible number formats.
	- Added `blurOnSelect` property to ArtifactSetAutocomplete for improved interaction.
	- Introduced a line break in the artifact editor layout.

- **Style**
	- Added a new styled line break component in the artifact editor interface.

### DIFF
--- a/libs/common/ui/src/components/NumberInputLazy.tsx
+++ b/libs/common/ui/src/components/NumberInputLazy.tsx
@@ -23,9 +23,10 @@ export function NumberInputLazy({
   useEffect(() => {
     setValue(valueProp?.toString())
   }, [valueProp])
+  const normalizedValue = float ? value.replace(',', '.') : value
 
   const saveValue = () => {
-    let num = float ? parseFloat(value) : parseInt(value)
+    let num = float ? parseFloat(normalizedValue) : parseInt(normalizedValue)
     if (isNaN(num)) {
       num = 0
       setValue(num.toString())
@@ -48,7 +49,7 @@ export function NumberInputLazy({
   ) => {
     const val = event.target.value
 
-    if (val.match(float ? /[^0-9.-]/ : /[^0-9-]/)) {
+    if (val.match(float ? /[^0-9.,-]/ : /[^0-9-]/)) {
       return event.preventDefault()
     }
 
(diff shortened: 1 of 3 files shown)


libs/common/ui/src/components/NumberInputLazy.tsx
```diff
@@ -23,9 +23,10 @@ export function NumberInputLazy({
   useEffect(() => {
     setValue(valueProp?.toString())
   }, [valueProp])
+  const normalizedValue = float ? value.replace(',', '.') : value
```
frzyc, line 26: Move this line into the `saveValue` function

libs/gi/ui/src/components/artifact/editor/index.tsx
```diff
@@ -649,6 +651,7 @@ export function ArtifactEditor({
                             >
                               {t('editor.uploadBtn')}
                             </Button>
+                            <LineBreak />
```
frzyc, line 654: Can you explain why this change fixes the undesired behaviour?

frzyc (changes requested):
Minor change + clarification needed

libs/gi/ui/src/components/artifact/editor/index.tsx
failchon: From my debugging, whenever the user focuses on the empty space or the space between the upload screenshot button and the ? button, the paste event listener detects that space as an input element. That input element comes from the right side of the modal, where you can input substats value.

My idea was to add an empty element that doesn't really affect the visuals or performance, so that the paste event listener targets that element instead of the input. It doesn't really have to be a br tag, but the other tags ( like divs ) brought other issues.

have not found a better solution to this, yet. So thought that a fix like this is helpful until we find a better solution.

frzyc: I appreciate your explanation. I am amidst implementing similar scanner feature to ZZZO, and knowing the "why" this behaviour happens is quite instrumental. Can you put a comment on this line with a lint to this PR/discussion?

failchon:
> Minor change + clarification needed

Resolved, please check again

frzyc (approved):
LGTM

PR #2654, opened 2025-01-30
ZO add advanced Set Filters

### LINKED ISSUE #2653: ZO Advanced set filters
Consider 4p/2p, rainbow and how they interact.
Use the filter to prefilter builds before the bulk of calcs

frzyc:
## Describe your changes
Adds a set filter:
- Default: `[]` any (any set/rainbow)
- User can quick select a set: `[setKey]` (disable rainbow, only builds with the set result)
- User can select multiple sets `[...setKey, ...]` (disable rainbow, only builds with those sets result)

Add a new stage in solver. Builds that fail the set filters gets eliminated. 
This should drastically increase build speed upon using the filter.

## Issue or discord link

- Resolves #2653

## Testing/validation
Using the "quick UI" to select a set
image
Using the "Advanced UI" to select multiple sets
image

### DIFF
--- a/libs/zzz/page-optimize/src/SetFilter.tsx
+++ b/libs/zzz/page-optimize/src/SetFilter.tsx
@@ -1,0 +1,246 @@
+import { useBoolState } from '@genshin-optimizer/common/react-util'
+import {
+  CardThemed,
+  DropdownButton,
+  ModalWrapper,
+} from '@genshin-optimizer/common/ui'
+import { toggleInArr } from '@genshin-optimizer/common/util'
+import type { DiscSetKey } from '@genshin-optimizer/zzz/consts'
+import { allDiscSetKeys } from '@genshin-optimizer/zzz/consts'
+import { useDatabaseContext } from '@genshin-optimizer/zzz/db-ui'
+import { DiscSetName } from '@genshin-optimizer/zzz/ui'
+import CloseIcon from '@mui/icons-material/Close'
+import {
+  Box,
+  Button,
+  ButtonGroup,
+  CardContent,
+  CardHeader,
+  Divider,
+  Grid,
+  IconButton,
+  MenuItem,
+  Typography,
+} from '@mui/material'
+import { useCharacterContext } from './CharacterContext'
+
+export function SetFilter({ disabled = false }: { disabled?: boolean }) {
+  const { database } = useDatabaseContext()
+  const [open, onOpen, onClose] = useBoolState()
+  const character = useCharacterContext()
+  const {
+    setFilter2 = [],
+    setFilter4 = [],
+    key: characterKey,
+  } = character ?? {}
+  return (
+    <CardThemed bgt="light">
+      <ModalWrapper open={open} onClose={onClose}>
+        <AdvSetFilterCard onClose={onClose} />
+      </ModalWrapper>
+      <CardContent sx={{ display: 'flex', gap: 1 }}>
+        <Box
+          sx={{ flexGrow: 1, display: 'flex', gap: 1, flexDirection: 'column' }}
+        >
+          {setFilter4.length <= 1 ? (
+            <DropdownButton
+              disabled={disabled}
+              title={
+                setFilter4[0] ? (
+                  <span>
+                    Force 4-Set:{' '}
+                    <strong>
+                      <DiscSetName setKey={setFilter4[0]} />
+                    </strong>
+                  </span>
+                ) : (
+                  'Select to force 4-Set'
+                )
+              }
+              sx={{ flexGrow: 1 }}
+            >
+              <MenuItem
+                onClick={() =>
+                  characterKey &&
+                  database.chars.set(characterKey, {
+                    setFilter4: [],
+                  })
+                }
+              >
+                No 4-Set
+              </MenuItem>
+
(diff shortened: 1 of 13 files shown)


libs/zzz/solver/src/calc.ts
```diff
@@ -13,22 +17,58 @@ import {
...
+      }
+    }
+    if (!has4) return false
+  }
+  if (filter2.length) {
+    for (const key in setCount) {
+      const k = key as DiscSetKey
+      const val = setCount[k]!
+      if (val === 1) return false // Rainbow
+      if (val >= 2 && val < 4 && !filter4.includes(k)) return false
```
nguyentvan7, line 45: logic looks wrong? why would we exclude a 2 or 3p based on filter4?

libs/zzz/solver/src/calc.ts
```diff
@@ -13,22 +17,58 @@ import {
...
+  filter2: DiscSetKey[],
+  filter4: DiscSetKey[]
+): boolean {
+  const setCount: Partial<Record<DiscSetKey, number>> = {}
+  if (!filter4.length && !filter2.length) return true
+  for (const { setKey } of discs) setCount[setKey] = (setCount[setKey] || 0) + 1
+
+  if (filter4.length) {
+    let has4 = false
+    for (const key in setCount) {
```
nguyentvan7, line 31: can do `Object.entries(setCount).forEach(([key, count]) => {})` and it will have proper typings and no need for `!` access

libs/zzz/solver/src/calc.ts
```diff
@@ -13,22 +17,58 @@ import {
...
+      const k = key as DiscSetKey
+      if (setCount[k]! >= 4 && filter4.includes(k)) {
+        has4 = true
```
nguyentvan7, line 34: nit: we seem to do more of arr.every() or arr.any() checks rather than this for-break paradigm. not a big deal tho, just weird to read

libs/zzz/solver/src/childWorker.ts
```diff
@@ -76,11 +84,15 @@ async function init({
...
   constraints: initCons,
   formulaKey: fk,
+  setFilter2,
+  setFilter4,
 }: ChildCommandInit) {
   baseStats = bs
   discStatsBySlot = discs
   constraints = initCons
   formulaKey = fk
+  SetFilter2p = setFilter2
```
nguyentvan7, line 94: capitalization

nguyentvan7 (changes requested):

libs/zzz/solver/src/calc.ts
frzyc: I use more for loops in calc because they are more performant (how much so I can't attest), I much rather use array function otherwise for robustness

libs/zzz/solver/src/calc.ts
frzyc: Yes, but its less performant, since it creates a new array

libs/zzz/page-optimize/src/SetFilter.tsx
```diff
@@ -0,0 +1,253 @@
...
+          color={
+            !setFilter4.length || setFilter4.includes(setKey)
+              ? 'success'
+              : 'secondary'
+          }
+          onClick={() =>
+            characterKey &&
+            database.chars.set(characterKey, {
+              setFilter4: setFilter4.length
+                ? toggleInArr([...setFilter4], setKey)
```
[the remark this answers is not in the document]
frzyc: In most cases, when you are selecting sets, you only really need to select 2-3 of them, so this is similiar to the behaviour of our main stat filter, where first click selects just that 1, which ends up being overall less clicks.

libs/zzz/page-optimize/src/SetFilter.tsx
```diff
@@ -0,0 +1,253 @@
...
+  CardContent,
+  CardHeader,
+  Divider,
+  Grid,
+  IconButton,
+  MenuItem,
+} from '@mui/material'
+import { useCharacterContext } from './CharacterContext'
+
+export function SetFilter({ disabled = false }: { disabled?: boolean }) {
```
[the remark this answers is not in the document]
frzyc: Its the same system, it allows people to "force" a single set without having to deal with the advanced config. I think quite a lot of the questions on GO was "how do you force a single set", now there is a shortcut without needing them to juggle a new concept

libs/zzz/solver/src/calc.ts
frzyc: good catch

libs/zzz/page-optimize/src/SetFilter.tsx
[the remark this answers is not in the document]
nguyentvan7: the main stat filter on ZZZ works like how i describe (only toggles the specific stat you clicked)

this just goes back to the 3-state button again though so i wont rehash it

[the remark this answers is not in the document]
frzyc: Yeah I will have to change the zenless one to be similiar to the 3 stat button. I am kinda annoyed the amount of clicks needed to set main stats

PR #2726, opened 2025-02-12
All: Change equipment on games to use special register function

nguyentvan7:
## Describe your changes

* Remove circular dependency in game-opt
* Create a shared registerEquipment function, based on previous gi-formula:registerArt function, which registers buffs in a special way for equipment to prevent high read traffic
* Update game formulas to use this paradigm
* Update debug filter for SRO
* Regenerate base sheets for SRO

## Issue or discord link

- 

## Testing/validation

Tests pass

Tested SRO conditionals and they still work as expected and formula debugging shows expected results


libs/game-opt/formula/src/util.ts
```diff
@@ -0,0 +1,59 @@
...
+  Sheet extends string
+>(
+  specificSheet: Sheet,
+  equipmentSheet: Sheet,
+  ...data: (
+    | TagMapNodeEntry<Tag_, Src, Dst, Sheet>
+    | TagMapNodeEntries<Tag_, Src, Dst, Sheet>
+  )[]
+): TagMapNodeEntries<Tag_, Src, Dst, Sheet> {
+  /* Unlike character and weapon, artifact buff is all-or-nothing, so we can register every
```
frzyc, line 21: Nit:Comments needs to be updated

libs/gi/formula/src/index.spec.ts
```diff
@@ -107,16 +112,15 @@ describe('sheet', () => {
...
           }
+          // eslint-disable-next-line no-fallthrough
           case 'enemy': {
             const desc = (enemyTag as any)[tag.qt]?.[tag.q]
             if (!desc) continue
             const { sheet } = desc
             if (!sheet) continue
             if (sheet === 'agg' || sheet === tag.sheet) continue
             fail(`Illform entry (${tagStr(tag)}) for sheet ${sheet}`)
-            break
```
frzyc, line 119: Would removing break here cause more issues since logic would go to the next case?

libs/sr/formula/src/data/lightCone/sheets/ASecretVow.ts
```diff
@@ -35,20 +34,25 @@ const sheet = register(
   registerBuff(
     'cond_dmg_',
     ownBuff.premod.dmg_.add(
-      boolConditional.ifOn(
-        cmpGE(lcCount, 1, subscript(superimpose, dm.cond_dmg_))
+      cmpGE(
```
frzyc, line 37: Is cmpGE more expensive than the conditional check? I wonder which one in the outside is better?

frzyc (commented):
Looks mostly good to me, some nit and questions that are non blocking

libs/sr/formula/src/data/lightCone/sheets/ASecretVow.ts
nguyentvan7: not sure if any is more expensive, but cmpGE outside allows better/easier debug filtering

libs/gi/formula/src/index.spec.ts
nguyentvan7: `fail` throws an error, so the break is unreachable and thus removed by linter. but then linter complains about fallthrough which does not exist

frzyc (approved):

PR #2866, opened 2025-03-18
ZO: Partitioned Char info into sections

failchon:
## Describe your changes

- Moved bonus stats, wengine conds and disc conds into the character info grid

## Testing/validation

image

## Summary by CodeRabbit

- **Refactor**
  - Simplified the bonus stats display by removing the header and divider elements.
  - Reorganized the character information layout by grouping related sections for improved clarity.

- **Style**
  - Updated card visuals with a refined outline and consistent layering for a more polished presentation.

- **New Features**
  - Introduced localization support for optimization features with a new JSON file containing English translations.


frzyc (changes requested):
I think the new sections are not "floating" or scrolling to the right place, i thought resetting the provider with a `0` value would fix it, but it seems that was not the fix. That's okay, since the sections was never designed to be nested like this, might need to cook a tad more.

(not for this PR) After changes in https://github.com/frzyc/genshin-optimizer/pull/2868, I think I want a unified conditional section, that only shows the currently equipped disc/wengine conditionals.

failchon:
> I think the new sections are not "floating" or scrolling to the right place, i thought resetting the provider with a `0` value would fix it, but it seems that was not the fix. That's okay, since the sections was never designed to be nested like this, might need to cook a tad more.
> 
> There is also excessive padding on the Bonus Stats section.
> 
> (not for this PR) After changes in #2868, I think I want a unified conditional section, that only shows the currently equipped disc/wengine conditionals.

Do we even want the scrolling behaviour for those sections? I kept them there but i don't think it's needed there.

Just a side note, this PR is not meant to do many css changes, but just to add the conds and bonus stats as part of this grid. More css changes would come after

frzyc (approved):

Issue #1498, opened 2024-02-13
Use Dialogs instead of Modals


frzyc:
A lot of prompts in GO would be better serviced with the [Dialog](https://mui.com/material-ui/react-dialog/) component.

However, there is sometimes an issue where the dialog backdrop will linger after its hidden, blocking interaction to the page:
`https://github.com/mui/material-ui/issues/32286`

We've fixed this for dropdown menus in the past: https://github.com/frzyc/genshin-optimizer/pull/967

But additional investigation is needed to see if the same solution can be applied to `Dialog`s. 

Currently, all dialog-able code should be tagged with `TODO: Dialog` in the comments.

PR #2929, opened 2025-04-04
Common: Fix invisible dropdowns once and for all

nguyentvan7:
## Describe your changes

* Fix invisible dropdowns stealing input once and for all, which should unblock #1498 

## Issue or discord link

- #936
- #1497

## Testing/validation

1. Open any TC build and change the goblet selector to a DMG bonus option
1. Open the DOM inspector and notice there is an invisible `div` left behind 
image
1. Anything under the invisible `div` should still be clickable

## Summary by CodeRabbit

- **Style**
	- Enhanced pointer event handling for overlay components, ensuring that elements not visibly rendered do not capture interactions.

### DIFF
--- a/libs/common/ui/src/theme/index.ts
+++ b/libs/common/ui/src/theme/index.ts
@@ -381,5 +381,23 @@ export const theme = createTheme({
         },
       },
     },
+    MuiPopper: {
+      styleOverrides: {
+        root: {
+          '&:has(> div[style*="opacity: 0"])': {
+            pointerEvents: 'none',
+          },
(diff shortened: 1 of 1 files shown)


frzyc (commented):
I think we've attempted this change before, this may some some weird interactions with closing out dropdowns in modals or something, I think we may need to examine counter cases to see this fixes all. Also, we are 2 major versions behind @mui/material(5.16.6 -> 7.0.1) so it may make sense for us to migrate to a new version before seeing if this problem persists.

PR #2945, opened 2025-04-17
ZO: Fix modal not opening when creating a new char on char page

failchon:
## Describe your changes

- Due to charOpt being a requirement for the modal to show, this wasn't happening because charOpt was not set on a new character that was added from character page. Useeffect was used her to add new char to charopt and also add existing chars to charopt ( if they aren't there already )

## Issue or discord link

https://discord.com/channels/785153694478893126/1353253559495299134

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability in loading character options, ensuring they are created or retrieved automatically when needed.

### DIFF
--- a/libs/zzz/page-characters/src/index.tsx
+++ b/libs/zzz/page-characters/src/index.tsx
@@ -102,7 +102,9 @@ export default function PageCharacter() {
     return characterKeyRaw as CharacterKey
   }, [characterKeyRaw, navigate])
   const character = useCharacter(characterKey ?? undefined)
-  const charOpt = useCharOpt(characterKey ?? undefined)
+  const charOpt =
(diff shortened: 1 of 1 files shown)


frzyc (changes requested):
This is a okay solution, but it will always have one render frame where you have no data
I would recommend implement it as:
`useCharOpt(characterKey ?? undefined) ?? (characterKey && database.charOpts.getOrCreate(characterKey))`

frzyc (commented):
Also, this probably is not going to be a long term solution, the real implementation is to implement this page without depending on charOpt. Once we hit M5, there won't be a singular charOpt for each character.

frzyc (approved):

PR #3077, opened 2025-10-03
Add better notificiation system

nguyentvan7:
## Describe your changes

Thanks Mimee

## Issue or discord link

- https://discord.com/channels/785153694478893126/1399530484312051854

### DIFF
--- a/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/index.tsx
+++ b/libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/index.tsx
@@ -306,6 +306,7 @@ export default function TabBuild() {
       maxBuildsToShow,
     } = buildSetting
     if (!characterKey || !optimizationTarget) return
+    Notification.requestPermission()
 
     const split = compactArtifacts(
       filteredArts,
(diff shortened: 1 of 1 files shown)


libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/index.tsx
```diff
@@ -306,6 +306,7 @@ export default function TabBuild() {
       maxBuildsToShow,
     } = buildSetting
     if (!characterKey || !optimizationTarget) return
+    Notification.requestPermission()
```
priolette, line 309: Just a question for my own interest, why do we have to call this request here and it's not enough to do it on the button click handler?

priolette (approved):
Looks good, notification works

libs/gi/page-team/src/CharacterDisplay/Tabs/TabOptimize/index.tsx
nguyentvan7: So that way if someone has the notification enabled, but permission not granted (they ignored it before, or had it enabled before we did this setup), when they click optimize, it prompts again. Removed the `await` so it doesn't block the optimization

PR #3176, opened 2026-03-10
Fix Lunar DMG Bonus constraint

Qm1719:
Adding { path } to the active node in nonStackBuff so that resolveInfo can then look up unit %, thus builds with ≥ X % Lunar DMG Bonus will now pass. Then, wrap the equal result with infoMut (already imported) to directly mutates matchNode.info = { path: 'lunarcharged_dmg_' }. Tested on a local version for every elements (non-moonsign char).

## Summary by CodeRabbit

* **Refactor**
  * Enhanced internal metadata tracking for improved data handling consistency.

### DIFF
--- a/libs/gi/sheets/src/SheetUtil.tsx
+++ b/libs/gi/sheets/src/SheetUtil.tsx
@@ -90,7 +90,7 @@ export function nonStackBuff(
   buffNode: NumNode | number
 ) {
   return [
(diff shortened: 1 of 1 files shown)


nguyentvan7 (approved):

nguyentvan7 (commented):
fix format issue (`yarn run biome check --fix`)

PR #3180, opened 2026-03-17
updating JSON import syntax

gabrielscn:
Summary

Replaces the deprecated JSON import assertion syntax with the newer import attributes syntax.

Changes

Updated:
import pkg from './package.json' assert { type: 'json' }
to:
import pkg from './package.json' with { type: 'json' }

Reason

The assert { type: 'json' } syntax is being replaced by the newer with { type: 'json' } syntax as part of the JavaScript import attributes specification.

## Summary by CodeRabbit

* **Style**
  * Updated JSON module import syntax in configuration.
  * Reformatted import spacing in the application file.

### DIFF
--- a/apps/zzz-frontend/src/app/App.tsx
+++ b/apps/zzz-frontend/src/app/App.tsx
@@ -2,6 +2,8 @@ import {
   AdBanner,
   AdBlockContextWrapper,
   AdRailSticky,
+
+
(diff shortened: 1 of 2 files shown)


nguyentvan7 (commented):
failed format, but looks good otherwise

PR #3253, opened 2026-07-09
Fix debug display using the wrong tag

nguyentvan7:
## Describe your changes

DebugDisplay was using Tag from TagContext, but we placed DebugDisplay at a very high level on the component stack. Further down the component stack, we will end up adding another TagContext layer, but the DebugDisplay would be unaware of that. To resolve that, we can just store the Tag into the DebugContext along with the Read

## Issue or discord link

- 

## Summary by CodeRabbit

- **New Features**
  - Formula debugging now also tracks and preserves the active tag when selecting a formula/field.
  - The debug modal opens only when both the selected read and tag are available.
- **Bug Fixes**
  - Closing the debug view now clears both the selected read and tag (not just the read).
  - Improved consistency so clicks in conditional, document, and field displays keep tag and formula details in sync.

### DIFF
--- a/libs/game-opt/formula-ui/src/DebugDisplay.tsx
+++ b/libs/game-opt/formula-ui/src/DebugDisplay.tsx
@@ -130,10 +130,9 @@ export function DebugListingsDisplay({
 }
 
 export function DebugReadModal() {
-  const tag = useContext(TagContext)
-  const calculator = useContext(CalcContext)?.withTag(tag)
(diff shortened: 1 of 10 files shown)


frzyc (commented):
the fix here is valid, and it addresses a real concern.

I wonder if there is a easier way to do this? 
why don't we wrap the modal body of the debug popup in `TagContextProvider value={capturedTag}`? then we only need to capture the intended tag onclick?

PR #3274, opened 2026-08-10
fix(zzz): Seed Besiege/Vanguard and teammate conditional toggles

IsPherz:
## Summary
- **Seed:** Add Vanguard `notOwnBuff` Besiege DMG%; keep Seed own Besiege as `ownBuff` (drop listing `true`). Vanguard ATK/CDMG gated by Attack + not Seed. Direct Strike is plain on/off; Vanguard is formula-selected.
- **Yanagi:** EX Special `anom_base_` → `teamBuff` (same polarity as Ult; removes `ownBuff` + `true` listing override).
- **Optimize teammate cards:** Lock writes to that teammate; coalesce legacy All/main rows; clear main then store `dst=null` on toggle.
- **Conditionals UI:** Remove src picker (src from TagContext / per-teammate card). Dst dropdown only when targeted with multiple targets.

## Test plan
- [ ] Optimize Harumasa with Seed + Astra: toggle Seed Direct Strike — Harumasa ATK/CRIT DMG move; Astra does not get Vanguard ATK
- [ ] With Direct Strike + Onslaught on: Besiege DMG% on both Seed and Harumasa
- [ ] Toggle Direct Strike off: leftover All/main rows don't keep the buff stuck
- [ ] Teammate conditionals show checkbox only (no src → dst chrome)
- [ ] Astra teammate conditionals still toggle and apply to main

### DIFF
--- a/libs/zzz/formula/src/data/char/sheets/Seed.ts
+++ b/libs/zzz/formula/src/data/char/sheets/Seed.ts
@@ -2,6 +2,7 @@ import type { NumNode } from '@genshin-optimizer/pando/engine'
 import {
   cmpEq,
   cmpGE,
+  cmpNE,
   prod,
   subscript,
   sum,
@@ -44,11 +45,13 @@ const { energy_consumed } = allNumConditionals(
   dm.m2.max_energy_consumed - dm.m2.energy_consumed
 )
 
+// Vanguard = Attack teammate, not Seed. Old check used team.initial.atk.max,
+// so Seed (also Attack) could be the max and block every ally from the buff.
 const directStrikeCheck = (node: NumNode | number) =>
   cmpGE(
     sum(
-      cmpEq(team.initial.atk.max, target.initial.atk, 1),
-      cmpEq(target.char.specialty, 'attack', 1)
+      cmpEq(target.char.specialty, 'attack', 1),
+      cmpNE(target.char.key, char.key, 1)
     ),
     2,
     node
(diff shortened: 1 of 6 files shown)


libs/zzz/formula/src/data/char/sheets/Seed.ts
```diff
@@ -44,11 +45,13 @@ const { energy_consumed } = allNumConditionals(
...
 const directStrikeCheck = (node: NumNode | number) =>
   cmpGE(
     sum(
-      cmpEq(team.initial.atk.max, target.initial.atk, 1),
-      cmpEq(target.char.specialty, 'attack', 1)
+      cmpEq(target.char.specialty, 'attack', 1),
+      cmpNE(target.char.key, char.key, 1)
     ),
     2,
     node
```
priolette, line 57: This doesn't handle the case if there are 3 attack characters in a team

libs/zzz/formula/src/data/char/sheets/Seed.ts
```diff
@@ -184,13 +187,18 @@ const sheet = register(
       )
     )
   ),
+  // Besiege DMG%: Seed (own) and Vanguard (notOwn), matching kit + M2 defIgn split.
```
priolette, line 190: what matching kit + M2 defIgn split?

priolette (changes requested):

libs/zzz/formula/src/data/char/sheets/Seed.ts
IsPherz: oh yes I dropped the max attack check cause seed was in theory able to meet that herself buit I should have keep the check and just exclude seed

libs/zzz/formula/src/data/char/sheets/Seed.ts
IsPherz: under besiege, seed and the vanguard both get the DMG% not the whole squad same pattern as M2 m2_defIgn_ is ownBuff (Seed) + m2_vanguard_defIgn_ is notOwnBuff so besiege common DMG% is split the same way

libs/zzz/formula/src/data/char/sheets/Seed.ts
IsPherz: kinda of ignored it cause we just reallty care forthe optimize target so I wasnt thinking of the status increase on another attack teammate

PR #3284, opened 2026-08-20
Compress extra DBs into b64-encoded gzip

nguyentvan7:
## Describe your changes

Both GO and ZO occupy the same limited 5 MB LocalStorage space. In addition, instant network scanners like Irminsul and zzz_packet_capture make it very easy for users to scan every single character, weapon and disc in an instant and import it to GO, resulting in larger databases than before. Compounding further, GO has been constantly adding features which further expand storage usage (teams, loadouts, multiopt, new artifact fields). Finally, the active DB for both sites must be flattened into LocalStorage for quicker reads/writes, causing this DB to be loaded into LocalStorage twice.

All of these combine to seeing more frequent reports of users missing data or having databases 'deleted' due to LocalStorage constraints. Mimee suggested compressing the JSON data before storing, and that seems to work great.

From the DBs I tested, there is ~90% reduction in storage usage for the compressed databases, which gives us substantial headroom.

There is also code in place to smoothly handle the transition. On first page load, all DBs will get loaded from LocalStorage in their stringified-JSON format, and then converted to compressed b64 gzip, so the storage savings will be automatic and immediate for users.

## Issue or discord link

- 

## Testing/validation

You can test on the PR site by swapping between the beta site, and the PR deployment. Going from beta -> PR deployment should preserve the database, and the storage usage in the Settings tab should drop.

## New Features

- Extra local database backups are now compressed, reducing browser storage usage.
- Database data is encoded for more efficient storage and recovery.
- Database initialization is handled consistently across the application.

## Bug Fixes

- Existing uncompressed database data remains supported.
- Compressed and legacy storage formats are restored automatically.
- Database state is restored more reliably across browser tabs and sessions.


libs/zzz/db-ui/src/provider/DatabaseProvider.tsx
```diff
@@ -17,7 +18,17 @@ export function DatabaseProvider({ children }: { children: ReactNode }) {
         return new ZzzDatabase(index, new DBLocalStorage(localStorage, 'zzz'))
       } else {
         const dbName = `zzz_extraDatabase_${index}`
-        const eDB = localStorage.getItem(dbName)
+        const rawDB = localStorage.getItem(dbName)
+        let eDB = ''
+        if (rawDB)
+          try {
```
frzyc: seems a dup of the same code in GO. should we hoist it as a common code handling? this also means we only handle GO/ZO, but not SRO, i think we shuold add it to SRO for parity.

apps/frontend/src/app/App.tsx
```diff
@@ -114,7 +115,17 @@ function App() {
...
+            // Handle if the DB is still an old uncompressed JSON string
+            JSON.parse(rawDB)
+            eDB = rawDB
+          } catch {
+            // If JSON parse fails, then it should be a compressed b64 gzip
+            eDB = unzipFromB64Gzip(rawDB)
+          }
         const dbObj = eDB ? JSON.parse(eDB) : {}
         const db = new ArtCharDatabase(index, new SandboxStorage(dbObj))
         db.toExtraLocalDB()
```
frzyc: Every init for inactive slots: read > decompress > parse > toExtraLocalDB() > recompress > write, even when already compressed. lets make sure this isnt run like every re-render or seomthing crazy. I don't think it is, but we shuold make sure.

apps/frontend/src/app/App.tsx
```diff
@@ -114,7 +115,17 @@ function App() {
...
       } else {
         const dbName = `extraDatabase_${index}`
-        const eDB = localStorage.getItem(dbName)
+        const rawDB = localStorage.getItem(dbName)
+        let eDB = ''
+        if (rawDB)
+          try {
+            // Handle if the DB is still an old uncompressed JSON string
+            JSON.parse(rawDB)
+            eDB = rawDB
```
frzyc: we `JSON.parse(rawDB)` for detection, and then later 
JSON parse again: `const dbObj = eDB ? JSON.parse(eDB) : {}`.

frzyc (commented):
compression is a good, easy win. i have some minor comments.

apps/frontend/src/app/App.tsx
nguyentvan7: I checked with console.log, looks to only do that full pass once a  full page load (e.g. f5 will trigger this once for every DB, swapping pages/tabs will not trigger it again)

frzyc (approved):

